PR #246, opened 2013-04-04
Add an 'all' argument to drilldown_tree_for_node

dracos:
This adds a new optional argument (immediately after the required arguments) to have the function return all descendants of the node, not just its immediate children.

I added this because I wanted the cumulative count functionality of drilldown_tree_for_node, but wanted to show all the descendants of a particular node, not just its children. I'm not sure if there's a different/better way I could have done this, feel free to let me know :)


craigds:
The template tags need a major update. Adding this as a keyword argument should be much more possible after #359.

matthiask:
Thanks. Please resubmit with resolved conflicts if you're still interested.

PR #258, opened 2013-06-10
templatetags: save level inside a data attr

xrmx:
Make it easier to get the node level from javascript.


matthiask:
Opinions, @craigds ? I like the approach of embedding additional attributes *inside* a field much better, especially because this still works if we remove the copy-pasted code in `mptt_admin.py` in the future.

(Example: https://github.com/matthiask/django-mptt/blob/mk/treeeditor/mptt/admin.py#L118)

Please reopen if you disagree (@xrmx too, please)

PR #265, opened 2013-07-10
Add node_moved signal to be sent after node successfully moved

gaqzi:
The signal behaves similar to the normal Django pre-/post_save signals.

The reason for this feature is that I'm writing a system which has inherited permissions. And when I move nodes around I need to re-check if permissions on descendants needs to be reset. The other solutions I tried wasn't working satisfactory and added quite a bit complexity. So by adding this to django-mptt I made the entire solution much cleaner.

I hope you will also find this useful or at least harmless enough so I don't have to maintain a fork. :)

### DIFF
--- a/docs/models.rst
+++ b/docs/models.rst
@@ -227,7 +227,8 @@ otherwise.
 -------------------------------------------
 
 Moves the model instance elsewhere in the tree based on ``target`` and
-``position`` (when appropriate).
+``position`` (when appropriate). If moved without any exceptions
+raised then the signal ``node_moved`` will be sent.
 
 .. note::
    It is assumed that when you call this method, the tree fields in the
--- a/mptt/managers.py
+++ b/mptt/managers.py
@@ -10,6 +10,7 @@ from django.db.models import F, Max, Q
 from django.utils.translation import ugettext as _
 
 from mptt.exceptions import CantDisableUpdates, InvalidMove
+from mptt.signals import node_moved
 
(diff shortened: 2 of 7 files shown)


matthiask:
Looks good to me. Any reason why this wasn't merged back then, @craigds ?

craigds:
no reason I can think of. Probably because I was busy at the time :(

This looks good, let's merge it

craigds:
I made some minor changes:

* fixed merge conflicts
* updated the mock_django dependency to >= 0.6.7 because otherwise we get errors on Python 3.
* made the dependency optional; tests that depend on it will just get skipped if it's not installed

Thanks :smile:

matthiask:
:+1: nice!

matthiask:
Under which circumstance would `self._base_manager` be set? The signal might be sent several times:

- `TreeManager.move_node` calls `TreeManager._move_node`
- `_move_node` calls `_base_manager.move_node`, which again calls `_base_manager._move_node`

`move_node` itself is called several times, therefore `node_moved` emitted several times. Shouldn't `_move_node` call `_base_manager._move_node` instead?

(Sorry — I'm don't know manager behavior too well.)

craigds:
Complicated. Yes, I think you're right.

To clarify, I think `_base_manager` is only set if the model is a subclass of a base which contains the tree fields (in multi-table inheritance)

Thanks for the heads up

PR #270, opened 2013-08-14
Basic support for InlineAdmins

jschrewe:
A very basic change that moves most of the reusable admin stuff into a mixin and adds inline admins. This is very basic but both changes might come in handy for someone.


matthiask:
Closing because of inactivity — please resubmit against current master, and include a simple test.

hoerin:
I wish the master supported InlineAdmins.

PR #322, opened 2014-08-26
Add support for models where primary key is not Integer type

ticosax:
Goal is to target models with UUID as primary key type.
See tests as an example.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,0 @@
-Django >= 1.4.2
-django_extensions


michaeljohnbarr:
@matthiask asked:
> So why is it necessary at all to introduce those get_db_prep_value calls?

django-extensions [UUIDField](https://github.com/django-extensions/django-extensions/blob/1.4.9/django_extensions/db/fields/__init__.py#L343) does not have a `to_python` method like Django's UUIDField does. Thus, when you call `node.pk` or `parent.pk`, it returns a string. Django's [`UUIDField.to_python`](https://github.com/django/django/blob/1.8a1/django/db/models/fields/__init__.py#L2378-L2388) returns a `uuid.UUID` instance.

When you pass a `uuid.UUID` instance as an argument to the `connection.cursor` as a list of params, it blows up, as outlined in #357. I am have coded a fix [HERE](https://github.com/michaeljohnbarr/django-mptt/blob/master/mptt/managers.py) which should address the issue, but I am trying to play nice and write some tests then commit a pull request. This (nearly) exact fix is running in my company's application right now (which uses a PostGres database), as MPTT is a critical piece of our application (thanks!).

Issue #257, opened 2013-06-10
Tree fields out of sync with the database, causes corrupt trees.


shelldweller:
Here's how to create a broken tree:

``` python
my_node = MyModel.objects.get(...)

# move things around in such a way that 
# my_node's MPTT attributes change,
# e.g., assign some of its children to another tree

my_other_node = MyModel.objects.get(...)
my_other_node.parent = my_node
my_other_node.save()
```

The tree is likely to be broken now. Most likely this happens because django-mptt takes parent's MPTT attributes at face value assuming they are in sync with its current db state.

On the other hand if in the above snippet parent assignment is substituted with `move_to`, tree does not break even if parent's mptt attributes are out of sync.

I suggest that parent's assignment should validate object's state, just as it probably does it for `move_to`.

yevhen-m:
@shelldweller I can see that `'parent'` is still present in `'update_fields'` even if unchanged. Why? This can cause problem in case parent was changed "outside" by someone else between the "read" and "write".

matthiask:
A recent commit fixed the most common stale-instance corruption scenario: when `save()` is called on a root node whose `order_insertion_by` field has changed, while other nodes' `tree_id`s have already been reshuffled since the instance was loaded, the raw SQL in `_make_sibling_of_root_node` would target the wrong row and `super().save()` would then write the wrong `tree_id` back, creating duplicates. The fix refreshes the four MPTT tree fields from the DB before executing the repositioning move (fixes #687).

For context on what remains:
- The `same_order = True` branch (plain field changes like `page.title = 'new'`) was already safe — it uses `update_fields` that excludes MPTT tree fields.
- `move_to()` does not go through `save()`, so stale instances before `move_to()` still require a manual `instance.refresh_from_db()` call first.
- Writes from outside MPTT (raw SQL, `bulk_update()`, etc.) cannot be defended against from MPTT's side.

PR #354, opened 2015-01-14
basic concurrency support (current master) #257

shelldweller:
Tested with Django 1.4, 1.6, 1.7 on Python 2.7 and 1.6, 1,7 on Python 3.3.


craigds:
This looks like an awesome start, probably fixes most of the concurrent-update issues people are having. Thanks!

PR #358, opened 2015-02-01
Treequeryset

karabijavad:


karabijavad:
This should allow for .get_ancestors() and .get_descendants() on query sets.

eg:
```
Category.objects.filter(...).filter(...).get_ancestors()
```
is now possible, as opposed to:
```
my_company.get_ancestors().filter(...)
```

This is beneficial, atleast to me, because now i do not need to iterate over my multiple objects if i want the ancestors of multiple objects. I can get all the ancestors/descendants in one query

thanks

musashiXXX:
Just curious, but have you tried ``get_queryset_ancestors`` and/or ``get_queryset_descendants``? For example:

    qs = Category.objects.filter(...).filter(...)
    Category.objects.get_queryset_ancestors(qs)
    Category.objects.get_queryset_descendants(qs)
    Category.objects.get_queryset_ancestors(qs, include_self = True)
    Category.objects.get_queryset_descendants(qs, include_self = True)

karabijavad:
ah, @musashiXXX thank you for that. that does help.

however, i believe that having the method in the queryset object would be nice...
is there a reason this isnt currently implemented?

craigds:
Thanks for the PR!

I don't have a concrete reason this isn't implemented. It does indeed make sense to have those methods in the queryset. I think the main reason is merely that I got the impression that the Django community seemed to frown upon custom querysets [until recently](https://docs.djangoproject.com/en/1.7/topics/db/managers/#creating-manager-with-queryset-methods), and all the queryset-ish logic was meant to go in the manager.

The implementation needs a bit of tweaking - it should probably do what `TreeManager.get_queryset_[descendants|ancestors]` does at the moment, and those manager functions should just call the queryset version.

karabijavad:
@craigds i agree, had i known about TreeManager.get_queryset_[descendants|ancestors] i would have actually just literally called that method

i will do that then push up again
thanks

karabijavad:
closed opened to trigger rebuild

the fail was anamolous, connection reset by peer when attempting to install a pip package.

@craigds i just did what you recommended, what do you think?

karabijavad:
sorry for the close reopen nonsense, but the travisci thing keeps saying its failing when its really passing...

mptt/managers.py
```diff
@@ -94,15 +95,13 @@ def get_query_set(self, *args, **kwargs):
 
         This method can be removed when support for Django < 1.6 is dropped.
         """
-        return super(TreeManager, self).get_query_set(*args, **kwargs).order_by(
-            self.tree_id_attr, self.left_attr)
+        return TreeQuerySet(self.model).order_by(self.tree_id_attr, self.left_attr)
```
craigds: I think this should use `TreeQuerySet(self.model, using=self._db)`... for multi-db support.

mptt/managers.py
```diff
@@ -94,15 +95,13 @@ def get_query_set(self, *args, **kwargs):
...
 
     def get_queryset(self, *args, **kwargs):
         """
         Ensures that this manager always returns nodes in tree order.
         """
-        return super(TreeManager, self).get_queryset(*args, **kwargs).order_by(
-            self.tree_id_attr, self.left_attr)
+        return TreeQuerySet(self.model).order_by(self.tree_id_attr, self.left_attr)
```
craigds: same here

PR #359, opened 2015-02-07
Modernise template tags

craigds:
WIP.

Brings template tags up to date with modern stuff (`@register.assignment_tag`, etc).

Requires some backwards incompatible changes (e.g. unquoted `myapp.modelname` not supported any more, but passing actual model classes will be.)

This PR creates a deprecation path for 0.7, for backward incompatible cleanup to go into 0.8.

* [x] `full_tree_for_model`
* [ ] `drilldown_tree_for_node`
* [ ] others?


craigds:
decided this isn't really that important, mostly aethestic changes, and it will cause backwards incompatibilities so I'm going to abandon it.

Issue #369, opened 2015-04-22
Using cls.__dict__ sometimes returns <class 'django.db.models.manager.ManagerDescriptor'>


benjaoming:
I'm having problems that I didn't get to the bottom of, except that reverting back to `getattr` instead of accessing via `__dict__` is the remedy.

Problem was that a custom manager was not set, and we were defaulting back to TreeManager.

The culprit was this:

                            # HACK: avoid using getattr(cls, attr)
                            # because it calls __get__ on descriptors, which can cause nasty side effects
                            # with more inconsiderate apps.
                            # (e.g. django-tagging's TagField is a descriptor which can do db queries on getattr)
                            # ( ref: http://stackoverflow.com/questions/27790344 )
                            obj = cls.__dict__[attr]

It was causing django-wiki/django-wiki#404

PR coming in......

benjaoming:
Here's the exact cause: a01b157ed15d412c664f31b5c0d3b4f3eddc0937

But I realize it fixes another important issue.

craigds:
Hi Ben

The last comment on the django-wiki ticket implies this is something to do with `.none()` querysets, but I'm not quite sure how that relates to `ManagerDescriptor`.

Perhaps you could you clarify a bit?

benjaoming:
@craigds thanks for getting back

I had to halt the investigation, but I'm trying to figure it out. I was building a test case for django-mptt but it somehow succeeds (oh no!!! `type(Model.objects.all()[0].get_children()) is CustomQuerySet`) with the current codebase, but the *exact* same thing in django-wiki is failing with the same structure (`type(Model.objects.all()[0].get_children()) is TreeQuerySet` instead of `CustomQuerySet`).

Human rant: So the problem caused by the way that django-wiki's custom manager... something isn't playing the right way, since it's returning a `ManagerDescriptor` (that's a class that will just raise an exception about not using manager functions directly on an object), and then django-mptt defaults to TreeManager because it doesn't find the custom manager inheriting from TreeManager.

benjaoming:
Btw thanks for a great project again!!!

benjaoming:
Okay I've found the problem, it's essentially this:

```python
class MyModel(MPTTModel):
    objects = MyCustomManager()
    _default_manager = objects  # <- THAT IS THE PROBLEM
```

benjaoming:
...now I just need to figure out why that line was introduced

benjaoming:
[...]

PR #370, opened 2015-04-22
Fix for #369

benjaoming:
Small optimization for the way that the TreeManager is found and add some extra tests

### DIFF
--- a/mptt/models.py
+++ b/mptt/models.py
@@ -327,21 +327,27 @@ class MPTTModelBase(ModelBase):
 
                 # make sure we have a tree manager somewhere
                 tree_manager = None
-                for attr in sorted(dir(cls)):
-                    try:
-                        # HACK: avoid using getattr(cls, attr)
-                        # because it calls __get__ on descriptors, which can cause nasty side effects
-                        # with more inconsiderate apps.
-                        # (e.g. django-tagging's TagField is a descriptor which can do db queries on getattr)
-                        # ( ref: http://stackoverflow.com/questions/27790344 )
-                        obj = cls.__dict__[attr]
-                    except KeyError:
-                        continue
-                    if isinstance(obj, TreeManager):
-                        tree_manager = obj
-                        # prefer any locally defined manager (i.e. keep going if not local)
-                        if obj.model is cls:
-                            break
+                attrs = dir(cls)
+                if "objects" in attrs and isinstance(cls.objects, TreeManager):
+                    tree_manager = cls.objects
+                
+                # Go look for it somewhere else
+                else:
+                    for attr in sorted(attrs):
+                        try:
+                            # HACK: avoid using getattr(cls, attr)
+                            # because it calls __get__ on descriptors, which can cause nasty side effects
+                            # with more inconsiderate apps.
+                            # (e.g. django-tagging's TagField is a descriptor which can do db queries on getattr)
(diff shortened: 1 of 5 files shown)


benjaoming:
Not sure why the doctests are only failing on some django versions? I've added a fix to a problem that ideally would have made all tests fail :)

benjaoming:
Sorry that this is getting confusing, just that I can't build all the environments locally, so having travis do the work is gr8 :+1:

craigds:
np, I do that too, since I still haven't managed to get homebrew to compile *any* version of python 3 on my laptop :(

benjaoming:
Okay after having uncommented the `_default_manager` part, it seems that the test that asserts this:

    self.assertTrue(isinstance(Person.objects.all()[0].get_children().none(), CustomTreeQueryset))

...which is a new test... is failing because of something different... I'll leave it out for this PR and then maybe we can find the separate cause of that?

Would that be okay?

benjaoming:
I mean, it's only failing in Django 1.4 so it's kind of not the end of the world :)

craigds:
I was under the impression that Django changed the behaviour of `.none()` recently? But I can't see that noted in [the docs](https://docs.djangoproject.com/en/1.8/ref/models/querysets/#django.db.models.query.QuerySet.none).

The docs say it should return an `EmptyQuerySet` so I don't really understand why you're asserting it's a `CustomTreeQueryset`.

benjaoming:
Adding issue #371 for the django 1.4 test failure which has then just been removed from the tests for now until further decision is made.

benjaoming:
Oh yes, you're right @craigds -- maybe we should create a test that asserts that it just has the custom methods of the custom queryset instead of looking at the type? Because AFAIK in Django 1.4, the approach would have been to create a `CustomEmptyQuerySet` as a separate type.

benjaoming:
okay now we're at it, I'll just add the modification in here with the other things... so it's like:

1. Fix stuff in case someone was using `_default_manager`
1. Optimize stuff by not iterating everything to find a TreeManager on a Model, just look at `objects`
1. Create more tests for checking out custom stuff
1. Create a test for the .none() method
1. Alterate the test Manager to reflect the way that managers are written differently depending on Django 1.4/1.5 og Django 1.6+

benjaoming:
Oh for crying out loud Travis, these tests work locally \*sigh\*

benjaoming:
I can't seem to get the last part to work, namely to make Django 1.4 find custom methods on a customized EmptyQuerySet returned by a custom Manager's `get_empty_query_set()` method... hence the failures.

So well okay, maybe it's not that important? Let me just remove `_default_manager = objects` again to see if that's the cause although I really doubt it...

craigds:
Yeah, I didn't think the empty queryset was guaranteed to have any custom methods on it. Maybe it does nowdays, but it doesn't surprise me that it doesn't in django 1.4.

It's probably fine to merge without that assert.

craigds:
Probably a bit late for this, but you can run `tox` to run the tests in a bunch of different environments. Or `tox -e py27-14` specifically for django 1.4

craigds:
Looks good. So this does what's required for django-wiki now? Do you mind squashing these into one commit? Cheers

benjaoming:
@craigds yeah that's what I did but everything just passed here locally, also the `.none()` on Django 1.4.

benjaoming:
Squashing and making new PR...

PR #389, opened 2015-07-08
Cleaned up tox.ini using new tox features.

spookylukey:
See https://tox.readthedocs.org/en/latest/config.html#generating-environments-conditional-settings
for more information


craigds:
That is ... awesome

PR #396, opened 2015-08-14
[IMPROVE]make get_queryset_ancestors/descendants SQL querys O(1)

yjmade:
before change, if query get n items, there will be n+2 SQL query, and each of them fetch all the fields of the query model but actually not being use at all.

After change, no matter how much items that query contains, there will be always one query and just fetch the lft,rght,and parent_id fields.


craigds:
merged via #403, thanks :)

Issue #335, opened 2014-10-11
raw_id_field and MPTTModelAdmin are uncompatible


krasnoperov:
It is not possible to use raw_id_fields together with MPTTModelAdmin.

Let's assume following model and admin:

    class Rubric(models.Model)
        parent = TreeForeignKey('self')

    class RubricAdmin(MPTTModelAdmin):
        raw_id_fields = ('parent', )

In Admin interface on Rubric Edit Page it is possible to open popup (by click on search icon near rubric raw_id field) and choose desired parent rubric. Normally, user clicks on name of selected item in popup and ID of that item appears in rubric field.

But actually nothing happens. Popup refreshed after click and that is all.

Reasons are simple. Item's link in popup must be like this:

    <a href="/admin/rubrics/rubric/1/?_changelist_filters=_popup%3D1%26_to_field%3Did" onclick="opener.dismissRelatedLookupPopup(window, '1'); return false;">item</a>

But actually '_changelist_filters' are missed:

    <a href="/admin/rubrics/rubric/1/" onclick="opener.dismissRelatedLookupPopup(window, 2'); return false;">item</a>

Any ideas how to fix it?

P.S. Version of Django is 1.7, django-mptt is 0.6.1 and master

matthiask:
Cross-posting https://github.com/django/django/commit/534ced5aadf964eca4cf29a689dc70185f582772 here as well.

PR #407, opened 2015-10-22
Fixed raw_id_field and MPTTModelAdmin are uncompatible

marctc:
Fixed raw_id_field and MPTTModelAdmin are uncompatible https://github.com/django-mptt/django-mptt/issues/335

### DIFF
--- a/mptt/templatetags/mptt_admin.py
+++ b/mptt/templatetags/mptt_admin.py
@@ -126,7 +126,7 @@ def mptt_items_for_result(cl, result, form):
             else:
                 attr = pk
             value = result.serializable_value(attr)
-            result_id = repr(force_text(value))[1:]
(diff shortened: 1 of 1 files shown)


craigds:
Could you explain this a bit? I don't really know what I'm looking at here

marctc:
TBH, I exactly don't know, but it really works.

matthiask:
We should probably switch to `escapejs()` instead of abusing `repr()` — Django did the same in 2013: https://github.com/django/django/commit/534ced5aadf964eca4cf29a689dc70185f582772

matthiask:
Thinking some more about this, the current code probably tries stripping the leading `u` from unicode strings. The proposed code strips quotation marks or something. 

This does not change the fact that using `escapejs` would still be better.

matthiask:
#429 instead. Thanks!

PR #410, opened 2015-10-31
remove support old django version

unehub:
Remove support old django versions and fixed problems with empty_value_display


craigds:
Thanks. Seems like the test failures are legitimate, so this'll need some fixing before I can merge it

craigds:
Done via 05e908d97b1fd3d1b563d18c4f95b58313b3fc27

Issue #413, opened 2015-11-20
Move cache_tree_children to TreeManager


danielnaab:
It would be great to be able to query a tree and get back an object that may be traversed using standard Python object notation. For example:

    # This (with the hypothetical use of `as_tree`) should result in only two database hits:
    root = TreeNode.objects.get(pk=1)
    tree_root = root.get_descendants(include_self=True).select_related('foreign_key').as_tree()
    for child in tree_root.children:
        print child.foreign_key.pk

If there a way to do this already I would be thrilled to know, as I need to deal with a tree among disparate pieces of code and it's not clear how to efficiently pass a node into a function that may traverse the tree without it triggering unnecessary DB hits.

I will create a separate package to do this if not a reasonable addition to the core library or there doesn't already exist a third-party page that does this. I'd appreciate some pointers on if that is the necessary path.

danielnaab:
I see that `mptt.templatetags.mptt_tags.cache_tree_children` does what I'm looking for, but isn't documented. I will create a pull request documenting its use, but where should it go? Maybe just a FAQ entry?

Also, an observation: since the `cache_tree_filter` is useful outside the context of templates, perhaps the core implementation should not be in `mptt_tags`, but instead on the model manager?

craigds:
Hi Daniel

Yes, moving it to the model manager would make sense (and log a deprecation warning for the templatetags one). Cheers

danielnaab:
Hi Craig,

Thanks, I can look into doing that.

One additional question: It seems that `get_ancestors()`, along with `get_children()` should honor the cache. `get_children()` works perfectly, but although I see code that looks like `get_ancestors()` does, my tests show that accessing attributes on the nodes returned by it results in a db hit. Are there unit tests to confirm this functionality? I'm not certain if I'm doing something wrong or not.

Thanks!

danielnaab:
FYI to anyone else who encounters this:

The issue I saw with extra DB hits was related to using methods on the related querysets, such as `get_ancestors().last()` or `get_children.all()`. Iterating over the results of `get_children()` or `get_ancestors()` directly doesn't result in db hits when using `cache_tree_children`.

craigds:
FWIW there is a test that covers `get_ancestors()` using the cache (`RecurseTreeTestCase.test_cached_ancestors`)

craigds:
Thanks :)

[...]

PR #416, opened 2015-12-02
Move `cache_tree_children` to `TreeQuerySet.get_cached_trees`.

danielnaab:
@craigds, for #413 it seemed most appropriate to put this on the `TreeQuerySet` rather than on the `Manager`, as it allows one to get an entire tree with one query (you don't need to query the root nodes first, then children). The existing code works in this context with very little change, and I left the current `cache_tree_children` tag in place with a `DeprecationWarning`.

Let me know how this looks, or if needs additional changes.


tests/myapp/tests.py
```diff
@@ -1806,3 +1806,51 @@ def test_insert_ordered_BFS_backwards_nonroot_nodes(self):
...
+
+        # Getting the ancestors of each node should not result in db hits.
+        with self.assertNumQueries(0):
+            self.assertEqual(len(root_nodes), 1)
+            category = root_nodes[0]
+            self.assertIsInstance(category, Category)
+            self.assertEqual(len(category.get_children()), 3)
+            for child in category.get_children():
+                self.assertIsInstance(child, Category)
+                for child in category.get_children():
```
craigds: was this meant to iterate over the children of `child` instead of `category`?

tests/myapp/tests.py
```diff
@@ -1806,3 +1806,51 @@ def test_insert_ordered_BFS_backwards_nonroot_nodes(self):
...
+        """
+        Test a query with one root node.
+        """
+        with self.assertNumQueries(1):
+            root_nodes = Category.objects.all().get_cached_trees()
+
+        # `get_cached_trees` should only return the root nodes
+        self.assertEqual(len(root_nodes), 1)
+
+        # Getting the ancestors of each node should not result in db hits.
```
craigds: Looks like this only tests `get_children()` but the comment says ancestors

tests/myapp/tests.py
```diff
@@ -1806,3 +1806,51 @@ def test_insert_ordered_BFS_backwards_nonroot_nodes(self):
...
+            self.assertEqual(len(root_nodes), 1)
+            category = root_nodes[0]
+            self.assertIsInstance(category, Category)
+            self.assertEqual(len(category.get_children()), 3)
+            for child in category.get_children():
+                self.assertIsInstance(child, Category)
+                for child in category.get_children():
+                    self.assertIsInstance(child, Category)
+
+    def test_genre_iter(self):
```
craigds: Does this test a different edge case than the other test?

If not, maybe just have one of them :)

craigds:
Thanks, awesome work! A couple of tweaks to the tests (as noted) and then I think I can merge this :smile:

craigds:
Looks great. One more thing: should probably add this method to the docs :)

danielnaab:
@craigds The latest commit has some docs. I also lightly filled in some bits that were missing elsewhere.

Issue #336, opened 2014-10-22
Draggable admin view


mazing:
Are there any plans to implement a draggable admin view so the structure can be changed easier?

It seems there are multiple solutions based on mptt, but some of them are a bit outdated and I think it would be a great implementation to the original project.

matthiask:
Thanks, @craigds. I haven't seen this one. I have been thinking about splitting FeinCMS into a tree editing component and a content editing component for quite some time now, but integrating a stripped down version of the TreeEditor into django-mptt sounds so much better to me. 

Of course I'd continue doing my part keeping the code up to date.

(Side note: django-treebeard already has a tree editing interface based on FeinCMS: https://tabo.pe/projects/django-treebeard/docs/2.0/admin.html#advanced-interface)

matthiask:
@mazing Have look at https://github.com/django-mptt/django-mptt/pull/426 if you find the time!

matthiask:
Fixed in https://github.com/django-mptt/django-mptt/commit/1041158647bb3c7ee84a3a0d6f04e65379f264de

PR #426, opened 2015-12-16
Initial working version of a stripped down FeinCMS TreeEditor 

matthiask:
Refs #336.

tree

Simple tests and some docs included. Comments and feedback very much welcome!

(And a happy new year to everyone :-)


matthiask:
@craigds Anything you want me to add, change, document, test? Would be great if you find the time for a review. I'm quite happy with things as they are now.

docs/admin.rst
```diff
@@ -42,3 +42,52 @@ to your MPTTModelAdmin::
...
+Usage::
+
+    from django.contrib import admin
+    from mptt.admin import DraggableMPTTAdmin
+    from myproject.myapp.models import Node
+
+    admin.site.register(
+        Node,
+        DraggableMPTTAdmin,
+        list_display=('__str__', ...more fields if you feel like it...),
```
craigds: I'd change this example to 'name' rather than '__str__' since I think the meaning would be clearer. (since models don't have a meaningful '__str__' method by default)

mptt/admin.py
```diff
@@ -84,3 +89,174 @@ def get_actions(self, request):
...
+
+        opts = self.model._meta
+        self.change_list_template = [
+            'admin/%s/%s/draggable_mptt_change_list.html' % (
+                opts.app_label, opts.object_name.lower()),
+            'admin/%s/draggable_mptt_change_list.html' % opts.app_label,
+            'admin/draggable_mptt_change_list.html',
+        ]
+
+        # Set a better header than "title"
```
craigds: So this sets the header of the first column to the name of the model?

It seems wrong to be changing `self.__class__...` in an `__init__` method. There must be a way to decorate the instance-bound method instead of the unbound function?

I think it should use the `short_description` of the first column name from `list_display`, rather than the name of the model. Then subclasses can override it easily.

docs/admin.rst
```diff
@@ -42,3 +42,52 @@ to your MPTTModelAdmin::
...
+    )
+
+Two columns are inserted after the action checkbox column with tree actions
+(move node, toggle subtrees) and an indented title according to the level of the
+tree node. The indented title takes its value from the model instance's
+``__str__`` method. The ``__str__`` entry itself is automatically removed
+from ``list_display``, but it is still recommended to leave it in there for
+clarity.
+
```
craigds: One thing that'd be useful to add here: what browsers does the javascript work on? I see a compat fix for IE<9 down there, does that mean it works in IE8?

craigds:
This looks amazing, thanks so much :) :trophy:

docs/admin.rst
matthiask: Thanks — the magic mechanism here was really hard to explain. Therefore I removed the magic in  https://github.com/django-mptt/django-mptt/commit/690e48d46ef92ef07b3b760f3fd60f51f9ad7d6d

mptt/admin.py
matthiask: I added an example to the docs of how `indented_title` might be overridden/replaced.

docs/admin.rst
matthiask: TBH I'm not sure myself. Will have to do some testing :-)

matthiask:
Thanks (not only but right now especially) for the review! I hope I addressed all your comments — certainly things are much clearer for it now.

matthiask:
Now with even less hardcoded values and even better documentation.

matthiask:
@craigds Any comments? Ready to go?

craigds:
Looks awesome, thanks again for doing this.

I think we still need a note about browser support. I'll give it a spin in a few things on Saucelabs today. What browsers/versions have you tried it in?

matthiask:
Thanks for reminding me twice!  :-(

Everything works fine in IE9 and better. I don't have a working MS Edge installation unfortunately, but since things work fine in Firefox, Chrome, Safari and IE9 I think we are safe.

IE7 fails because it does not know about `JSON.parse`. I don't think polyfilling is worth it given the following support matrix: http://caniuse.com/#search=json

matthiask:
Two polyfills were required for the following screen:

ie7-mptt

JSON: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON#Polyfill
sessionStorage: https://gist.github.com/tagawa/2880273

Looks a bit ugly but works fine. What do you think?

matthiask:
This branch https://github.com/matthiask/django-mptt/tree/mk/treeeditor-polyfills includes the polyfill commit https://github.com/matthiask/django-mptt/commit/70a2094b6fed764f8c56f0e5c8f319bd37f09ac6

matthiask:
One more thing concerning browser support: Django 1.9 does not support IE8 and below anymore — https://docs.djangoproject.com/en/1.9/releases/1.9/#contrib-admin-browser-support

craigds:
@matthiask I'm not particularly fussed about supporting old browsers. I just wanted a note in the docs to say which ones work :) IE9+ is fine. Definitely let's not support IE7, especially if it already looks like that. I haven't supported IE7 for anything for a while, and I don't particularly miss it ;)

craigds:
Go ahead and merge this if you're happy with it. Thanks a lot!

matthiask:
Thank you! Let's go :-)

ticosax:
:+1: Great work here ! I'm looking forward for the next release.
@matthiask Thank you for this.

Issue #107, opened 2010-12-19
Make MPTTModel.tree the default manager


craigds:
As far as I can tell there's no reason why the TreeManager is not the default manager. This confuses me and probably everyone else.

For 0.5, I think .objects should refer to the same manager as .tree . It can be overridden if a subclass wants a different default manager.

slenart:
FYI, the new Django can't set _default_manager, and actually it was bad practice to overwrite it.

PR #427, opened 2015-12-19
Use ._tree_manager in MPTT's own code instead of relying on .objects

matthiask:
I'm a bit surprised that `_tree_manager` isn't mentioned anywhere in the documentation. 

I know (and wholeheartedly approve) about #107 / https://github.com/django-mptt/django-mptt/commit/1651bf75d75af8c17a89bfd02754e4d85f84e658, but I'm still somewhat unsure what the "official way" to access the tree manager would be. Or should be always use `_default_manager` as is already the case in some parts of django-mptt's code?

### DIFF
--- a/mptt/admin.py
+++ b/mptt/admin.py
@@ -60,7 +60,7 @@ class MPTTModelAdmin(ModelAdmin):
         # If this is True, the confirmation page has been displayed
         if request.POST.get('post'):
             n = 0
-            with queryset.model.objects.delay_mptt_updates():
+            with queryset.model._tree_manager.delay_mptt_updates():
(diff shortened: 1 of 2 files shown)


craigds:
Thanks, these should definitely be `_tree_manager`.

Are you saying that you think there should be a way to explicitly access a tree manager from a third-party app (ie a public version of `_tree_manager`)?

matthiask:
I had the impression that Django documented the presence of `_default_manager` and that we could do the same with `_tree_manager`. That is not the case though — `_default_manager` is only mentioned once in the 1.8 release notes. Everything's fine :-)

PR #430, opened 2016-01-25
Fix #299. Selecting parent in list-filter field doesn't show children.

### LINKED ISSUE #299: Selecting parent in list-filter field doesn't show children
Models:
```
Material(MPTTModel):
    name = models.CharField(max_length=63)
    [mptt stuff, parent, order etc.]

Thing(models.Model):
    name = models.CharField(max_length=63)
    material = TreeForeignKey(Material)
```
Material data:
```
--- ParentMaterial
 | --- MaterialA
 | --- MaterialB
--- AnotherMaterial
[...]
```
There are `Thing` items with foreign keys pointing to MaterialA and MaterialB. In admin.py -> `ThingAdmin`, `material` is defined in `list_filter`. Choosing either MaterialA or MaterialB works fine to filter the items, but when ParentMaterial is chosen, nothing is shown as there are no items directly linking to it.

Is this expected behavior?

poleha:


craigds:
Sorry, I must have missed this a month ago.

This looks great. I'm having a bit of trouble wrapping my head around it though. Could you add some docs? I think a partial screenshot showing the filter would be particularly helpful.

poleha:
That's ok. I updated my PR with current master.
Please take a look at [screenshot](http://i.imgur.com/7dadec6.png).
I provided some inline docs:
```
class TreeRelatedFieldListFilter(RelatedFieldListFilter):
    """
    Admin filter class which filters models related to parent model with all it's descendants.

     Usage:

    from mptt.filters import TreeRelatedFieldListFilter

    @admin.register(models.MyModel)
    class MyModelAdmin(admin.ModelAdmin):
        model = models.MyModel
        list_filter = (
            ('my_related_model', TreeRelatedFieldListFilter),
                    )
    """
```
Since I just reuse django `RelatedFieldListFilter` and usage of `list_filter` is described in django [documentation](https://docs.djangoproject.com/en/1.9/ref/contrib/admin/#django.contrib.admin.ModelAdmin.list_filter) I thought that would be enough. Should I provide more information in documentation?

craigds:
Thanks :)

* I'd put it in `admin.py` since it's admin-related. That makes it clearer this is an admin filter, not some kind of database filter etc.
* The doc file isn't getting included in the docs build (`cd docs ; make html`), probably because it's not mentioned in `docs/mptt.rst`. Though if you move the code into `admin.py` it'll fix that anyway.
* I think a very brief description in `admin.rst` along a cropped version of that screenshot would be really helpful. Admin list filters are an advanced enough concept that having an image would really help to explain this. There aren't any images in the docs yet but it sounds like they're easy to add: http://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html#include-images

mptt/filters.py
```diff
@@ -0,0 +1,116 @@
...
+from django.contrib.admin.options import IncorrectLookupParameters
+from django.contrib.admin.utils import get_model_from_relation
+from django.core.exceptions import ValidationError
+from django.utils.encoding import smart_text
+from django.utils.html import mark_safe
+from django.utils.translation import ugettext_lazy as _
+from django.utils.translation import get_language_bidi
+from django.db.models.fields.related import ForeignObjectRel, ManyToManyField
+
+MPTT_ADMIN_LEVEL_INDENT = getattr(settings, 'MPTT_ADMIN_LEVEL_INDENT', 10)
```
craigds: I don't think this is required as a setting. Maybe just put it as a class attribute, and if anyone _really_ needs to override it they can subclass?

mptt/filters.py
```diff
@@ -0,0 +1,116 @@
...
+        model = models.MyModel
+        list_filter = (
+            ('my_related_model', TreeRelatedFieldListFilter),
+                    )
+    """
+    template = 'admin/mptt_filter.html'
+
+    # Ripped from contrib.admin.filters,RelatedFieldListFilter Django 1.8 to
+    # change lookup name to 'inhierarchy'
+    def __init__(self, field, request, params, model, model_admin, field_path):
```
craigds: Could this method just call super(...) and then override `lookup_kwarg`, rather than copying the rest of the method?

mptt/filters.py
```diff
@@ -0,0 +1,116 @@
...
+        self.title = self.lookup_title
+
+    # Ripped from contrib.admin.filters,FieldListFilter Django 1.8 to deal with
+    # lookup name 'inhierarchy'
+    def queryset(self, request, queryset):
+        try:
+            # #### MPTT ADDITION START
+            if self.lookup_val:
+                other_model = self.other_model.objects.get(pk=self.lookup_val)
+                other_models = other_model.get_descendants(True)
```
craigds: Since this is using `get_descendants`, does this mean that when you click on 'ChildMaterial1' the filter will _only_ show 'ChildMaterial1' ?

Would using `get_family` be better? Then it would show 'ParentMaterial' too so you could traverse back up the tree easily.

craigds: Ah, ignore this. This method produces the filtered queryset of related models, not the choices of the tree model

craigds:
Thanks a lot for this. I've just taken the time to actually figure out whats going on (I'm not very familiar with custom admin filters), and the code looks excellent. The tests in particular look very detailed :)

poleha:
Thank you for detailed review. I think that I fixed all the issues. I am not sure I could explain it well in documentation so constructive criticism would be appreciated.

Issue #434, opened 2016-02-16
DraggableMPTTAdmin improvements / setup.py troubles


matthiask:
Collecting a few problems with the current implementation of DraggableMPTTAdmin here (I'm working on those right now)

- [x] Make the implementation CSP-compatible (no inline JavaScript)
- [x] Remove the customized base template — use stock `admin/change_list.html` instead
- [x] When using filters, the `div#module.changelist` should also have the `filtered` class, otherwise filters overlap the change list area and hide cells beneath it

darkowic:
@matthiask what will happened with this? Will you continue it?

matthiask:
@darkowic Even better — the changes mentioned above have already hit master. I just forgot to mark the individual items as done!

darkowic:
Nice! I used old library with move functionality from feincms https://bitbucket.org/san/django-mptt-tree-editor and now I want to upgrade django but this is very nice feature :) Do you know when will it be released?

matthiask:
I don't know, but maybe @craigds can shed some light on this matter? We don't have any backwards-incompatible changes since 0.8, so maybe a release could be made quickly?

craigds:
I don't have any particular plans for 0.9, and no objection to releasing this in principle.

Though 0.8 is only 3 months old, and I've been sticking to supporting the last two minor releases. So releasing 0.9 would mean stopping supporting 0.7, which I'm not really happy with doing yet.

I suppose I could release master as 0.8.1 given that there's no backward-incompatible changes. This draggable admin is a substantial win, so it'd be a shame to wait ages for users to get it.

matthiask:
Great idea! Didn't think of that.

craigds:
Done :)

matthiask:
Awesome! Thank you :)

On Fri, Feb 26, 2016 at 11:27 PM, Craig de Stigter <notifications@github.com
> wrote:

> Done :)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/django-mptt/django-mptt/issues/434#issuecomment-189508661>
> .
>



-- 
Matthias Kestenholz
Partner, Programmierung

FEINHEIT AG
Molkenstrasse 21
8004 Zürich
www.feinheit.ch

darkowic:
Thank you guys! You are awesome!

mpachas:
@craigds Hi! I've just tried pypi install of 0.8.1 but it comes with no static folder, so I have both 404:
```
[...]

PR #437, opened 2016-02-16
Refs #434: Remove the need for custom draggable changelist templates

matthiask:
(Sorry for the spam!)


mptt/admin.py
```diff
@@ -92,28 +94,44 @@ def get_actions(self, request):
...
+    def __init__(self, js, attrs):
+        self.js = js
+        self.attrs = attrs
+
+    def startswith(self, _):
+        # Masquerade as absolute path so that we are returned as-is.
+        return True
+
+    def __html__(self):
+        return format_html(
```
craigds: Can you add a comment with an example of the output of this?

matthiask: Of course! 9dafd7686c729bf36b49f22231ea7cd55bc01ba3

This hack should work really well also with prehistoric versions of Django, back to https://github.com/django/django/commit/a19ed8aea395e8e07164ff7d85bd7dff2f24edca (the newforms-admin merge). The behavior of returning `TemplateResponse` instances from `changelist_view` was introduced with https://github.com/django/django/commit/c8092b840b4151724a6babdbeb84cf3138057c6a in 2011. 

(Adding data attributes to script tags is not my own idea btw, I got it here: https://github.com/django/django/blob/master/django/contrib/admin/templates/admin/popup_response.html)

craigds:
:+1: cheers, merge away when ready

Issue #243, opened 2013-03-31
Support for order_insertion_by on an update()


kirpit:
I was trying to use `order_insertion_by` with the combination of Django's order-with-respect-to feature by setting `order_insertion_by = ('_order', )`.

https://docs.djangoproject.com/en/dev/ref/models/options/#order-with-respect-to

When saving new objects on insert, django-mptt gives error because there is no `_order` field set yet but this can be hacked by a small custom save() method:

```
def save(self, *args, **kwargs):
    # set page order for new objects
    if self._order is None:
        parent = getattr(self, '_parent_cache', None) or self.parent
        self._order = Page.objects.filter(parent=parent).count()
    return super(Page, self).save(*args, **kwargs)
```

However, the problem arises if you use something like `obj.set_answer_order([3, 1, 2])` as it actually `update()`s the children, not `save()`s them. Looping through the children objects and explicitly saving them with `force_update=True` doesn't update either as there is some internal `same_order` mechanism (which I believe unnecessarily complicated) determining either to save or not, if any of the fields has been modified. Anyways, re-saving one-by-one shouldn't be the way of re-sorting..

Perhaps, we will face the same issue if we `update()` an object's field that is defined in the `order_insertion_by` such as `obj.update(slug='the-new-slug')` wont get sorted respectively.

matthiask:
See #833 — I want to keep the project working for the time being but I'm not going to add (and maintain) new features. Thanks for your understanding.

PR #443, opened 2016-03-11
#243 Support for order_insertion_by on an update()

poleha:


matthiask:
Thank you! I'm not completely sure what this change achieves. Can you provide some justification and/or an use case?

We already have ordering information, and if this is purely a different way of saying `order_insertion_by` I'm against it, but I suspect there's a good reason I'm not seeing (yet!)

poleha:
@matthiask, thank you. Please take a look at the [docs](https://docs.djangoproject.com/en/dev/ref/models/options/#order-with-respect-to). That's really not much to read and I don't think it's necessary to copy it here.
That's not just the different way of saying `order_insertion_by`.
As example, this django feature provides `set_RELATED_order` and `get_RELATED_order` methods.
It allows to explicitly set ordering of related children.
I suppose this feature should be added for two reasons:
1) It sometimes may be useful;
2) It's django feature, so I guess there's no reason for django-mptt to not support it.

matthiask:
Closing because of conflicts. Please resubmit if you still want to pursue this.

That being said, just because Django supports a particular feature does not mean that django-mptt necessarily also has to submit every variant too (limited maintainer time and attention etc.)

PR #444, opened 2016-03-13
Efficient bulk tree creation.

### LINKED ISSUE #442: Question / feature request: Bulk create tree from root node
As far as I can tell from the docs and the source, inserting nodes with tree properties (`lft`, `rght`, etc.) requires multiple database queries, even when passing flags like `save=False`. If a user wants to bulk-insert a large number of nodes on a tree that doesn't yet exist, it seems like they have to disable mptt, insert the nodes, and then rebuild the relevant partial tree. Instead, it should be possible to set all tree properties in python and do a single bulk insert. This would only work when the user wants to bulk create nodes in a tree that doesn't exist yet, so that it's safe to skip calling `_create_space` and other helpers that move around existing nodes. I'm imagining an interface like this:

``` python
nodes = [Node(**obj) for obj in objs]
set_tree_properties(nodes)
Node.manager.bulk_create(nodes)
```

Where `set_tree_properties` sets `tree_id`, `lft`, and `rght` on all nodes without touching the database. The idea is to create an entire tree with a single request, without the need for a potentially expensive `rebuild`.

Does this make sense? Would there be interest in adding a feature like this? I wrote a very simple implementation of `set_tree_properties` for a project I'm working on that I'd be happy to contribute, if it would be worthwhile.

jmcarp:
Allow bulk-creating a new tree or subtree while setting tree properties
in memory to minimize database queries.

[Resolves #442]

@craigds @erez-o: let me know if the proposed interface makes sense--I'll be happy to revise if you have suggestions.

cc @cmc333333


jmcarp:
Ping--let me know if you're interested in this feature. Feedback welcome!

matthiask:
I like it. The code looks fine, the tests sufficient.

craigds:
This sounds great :) Though I have a few questions:

 * should `bulk_load` be on the manager instead of the model class?
 * The items in the tests have `id` values already. Do the nodes have to have known IDs?
 * The nodes in the tests are all related to each other. Could I use this to create completely unrelated nodes, sprinkled arbitrarily across my forest?
 * is there a way we can use existing django interfaces for this (`Manager.bulk_create`, specifically) instead of a new method with new semantics? 

Re that last point, do you think there's a way we could make this interface work?

```
a = MyNode(name='a')
b = MyNode(name='b', parent=a)
c  = MyNode(name='c')
MyNode.objects.bulk_create([a, b, c])
```

Thanks for the effort you've put into this :)

jmcarp:
> should bulk_load be on the manager instead of the model class?
Now that you mention it, that sounds like the right place for it. Will move.

> The items in the tests have id values already. Do the nodes have to have known IDs?
The nodes don't have to have primary keys, but if a user wants to use `bulk_create` with foreign keys, primary keys should be pre-populated before save. If we need to use auto-incrementing keys, we need to run multiple `INSERT` queries, so `bulk_create` becomes less useful.

> Could I use this to create completely unrelated nodes, sprinkled arbitrarily across my forest?
My intention was to speed up the specific use case of inserting a tree of related nodes. For inserting a bunch of unrelated nodes, I think turning off MPTT updates, running `bulk_create`, and rebuilding is probably the best option.

> is there a way we can use existing django interfaces for this (Manager.bulk_create, specifically)
Not sure--I don't think overriding `bulk_create` is right, since `bulk_load` doesn't actually run an `INSERT`, and since I think we still want to expose the original `bulk_create` method, as in the use case above when we're inserting unrelated nodes.

Basically, the proposed interface works because it's constrained. I haven't figured out how to write an generic and efficient `bulk_create` that's better than `disable_mptt_updates` and `rebuild`.

matthiask:
@jmcarp Are you still interested resp. do you still have the time to work on this?

jmcarp:
@matthiask: sorry, this fell off my radar, but I'm definitely still interested in getting this in good shape. I just moved the `bulk_load` method from `MPTTModel` to `TreeManager` as suggested, and fixed the merge conflict. Let me know if this needs more revisions.

eressler:
Any update on when this will make it in? Even with disable_mptt_updates() importing very large trees takes hours (nine hours and counting on my current import)

jmcarp:
Bump. I'm happy to revise if there's anything missing.

mptt/managers.py
```diff
@@ -672,6 +672,79 @@ def partial_rebuild(self, tree_id):
 
         self._rebuild_helper(pks[0], 1, tree_id)
 
+    @delegate_manager
+    def bulk_load(self, data, target=None, position='last-child'):
```
craigds: From a cursory glance this method doesn't appear to support trees that use `order_insertion_by`, since it always appends nodes to the next available `tree_id`. Is that correct?

I imagine doing that would be much trickier. At a minimum I think this method should throw an error if you try to call it when using `order_insertion_by`

craigds: Not convinced about the name - I think it's too close to `bulk_create` and it's not obvious that this is a special tree-related thing. If I was tab-completing in a shell I'd be unsure which one to pick.

Maybe something like:
* `build_tree_nodes`
* `load_tree_nodes`
* `tree_bulk_load`

murdav:
Dear All, 
What is the status of this great/lovely/excellent bulk_load?
@craigds when do you program to merge it?

Thanks!

cirotix:
Hi @jmcarp @craigds,

I am definitively interested to get this merged. As I need to be able to `bulk_create` efficiently for my project, I will fork and apply the patch from @jmcarp on my branch. 

As this PR seems stalled, I can do  the changes asked by @craigds and open a new PR

mptt/managers.py
```diff
@@ -672,6 +672,79 @@ def partial_rebuild(self, tree_id):
...
+                else:
+                    cursor = getattr(target, opts.right_attr)
+        else:
+            tree_id = self.tree_model._tree_manager._get_next_tree_id()
+            cursor = 1
+            level = 0
+
+        stack = []
+
+        def treeify(data, cursor=1, level=0):
```
cirotix: Actually the `treeify` function uses the `stack` and `opts` variables from the enclosing scope so it should stay as it.

Tyler-Churchill:
Just discovered this project - looks awesome! I think this feature is a must, Any status on when this will be looked at/considered again for merge?

matthiask:
Closed in favour of https://github.com/django-mptt/django-mptt/pull/575

PR #455, opened 2016-04-22
Improve rebuild perfomance

un1t:


matthiask:
Hmm. The docstring of disable_mptt_updates recommends/mentions running rebuild inside a transaction (https://github.com/django-mptt/django-mptt/blob/ef523ad4162171e3bacca8ac5ab85ac1881b0d85/mptt/managers.py#L263)

If people followed the docs and there's an additional atomic() somewhere, I fear that if the rebuild fails people are left with a silently corrupted tree. I've added the recommendation to run rebuild inside atomic to the docs (d1c00a68cb06594b7a6e93af50f88e0c65a32c57).

(@craigds No objection from my side if you think this change is a good idea.)

PR #458, opened 2016-05-04
Add django master in TravisCI

ad-m:
Hello,

I am suggest add tests over django master to fast feedback what can be broken in future django releases. It is failurable tests, so provide feedback, but isn't misleading for developer which create a pull requests. It doesn't affect GitHub pull requests status.

I drop duplicated ```sudo: false``` too.

Greetings,

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,10 @@ matrix:
       python: "3.2"
     - env: DJANGO="Django>=1.9,<1.10"
       python: "3.3"
+    - env: DJANGO="https://github.com/django/django/archive/master.tar.gz"
+      python: "3.2"
+    - env: DJANGO="https://github.com/django/django/archive/master.tar.gz"
+      python: "3.3"
   allow_failures:
     - env: DJANGO="https://github.com/django/django/archive/master.tar.gz"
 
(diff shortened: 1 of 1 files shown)


claudep:
+1, the failures of the PR already caused a fix in Django (https://github.com/django/django/commit/3204bc8e5e4845cf9fbd12246f83ae851081870e)!

.travis.yml
```diff
@@ -18,11 +22,10 @@ matrix:
...
+  allow_failures:
+    - env: DJANGO="https://github.com/django/django/archive/master.tar.gz"
 
 # command to install dependencies, e.g. pip install -r requirements.txt --use-mirrors
 install: pip install $DJANGO mock-django==0.6.9
 # command to run tests, e.g. python setup.py test
 script: cd tests/ && ./runtests.sh
-
-# Run on travis' container infrastructure
-sudo: false
```
craigds: Why's this removed?

craigds: ah, duplicate. 👍

craigds:
Looks like this runs tests on python 3.2 & 3.3, which django master isn't compatible with. Can you add excludes for those (as with 1.9)?

ad-m:
Changed as you expect.

Issue #459, opened 2016-05-04
"Model" object has no attribute '_deferred` - django 1.10 compatibility


ad-m:
Hello,

Django 1.10 ( https://docs.djangoproject.com/en/dev/releases/1.10/ ) said:
> The Model._deferred attribute is removed as dynamic model classes when using QuerySet.defer() and only() is removed.

I believe this make django 1.10 & django-mptt following tests fail:
```pytb
======================================================================
ERROR: setUpClass (tests.myapp.tests.AdminBatch)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/django-mptt-env/local/lib/python2.7/site-packages/django/test/testcases.py", line 1019, in setUpClass
    'database': db_name,
  File "/tmp/django-mptt-env/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 130, in call_command
    return command.execute(*args, **defaults)
  File "/tmp/django-mptt-env/local/lib/python2.7/site-packages/django/core/management/base.py", line 356, in execute
    output = self.handle(*args, **options)
  File "/tmp/django-mptt-env/local/lib/python2.7/site-packages/django/core/management/commands/loaddata.py", line 64, in handle
    self.loaddata(fixture_labels)
  File "/tmp/django-mptt-env/local/lib/python2.7/site-packages/django/core/management/commands/loaddata.py", line 104, in loaddata
    self.load_label(fixture_label)
  File "/tmp/django-mptt-env/local/lib/python2.7/site-packages/django/core/management/commands/loaddata.py", line 161, in load_label
    for obj in objects:
  File "/tmp/django-mptt-env/local/lib/python2.7/site-packages/django/core/serializers/json.py", line 86, in Deserializer
    six.reraise(DeserializationError, DeserializationError(e), sys.exc_info()[2])
  File "/tmp/django-mptt-env/local/lib/python2.7/site-packages/django/core/serializers/json.py", line 80, in Deserializer
    for obj in PythonDeserializer(objects, **options):
  File "/tmp/django-mptt-env/local/lib/python2.7/site-packages/django/core/serializers/python.py", line 183, in Deserializer
    obj = base.build_instance(Model, data, db)
  File "/tmp/django-mptt-env/local/lib/python2.7/site-packages/django/core/serializers/base.py", line 218, in build_instance
    obj = Model(**data)
  File "../mptt/models.py", line 390, in __init__
    self._mptt_meta.update_mptt_cached_fields(self)
[...]

PR #461, opened 2016-05-05
Stop using Model._deferred private attribute

claudep:
That attribute has been removed in Django 1.10.
See https://github.com/django/django/commit/7f51876f9


claudep:
Fixes #459.

PR #463, opened 2016-05-05
Provide for the new 'required' input attribute in Django 1.10

claudep:


claudep:
I had a superfluous space...

claudep:
@craigds Any hurdle for this patch to be committed?

craigds:
No hurdles :) thanks for helping out!

claudep:
Thanks for merging, Craig. The follow-up would be to merge #458. It never ends :-)

Issue #357, opened 2015-01-23
TreeManager._move_child_within_tree breaks with UUIDField() as primary key (ID)


michaeljohnbarr:
Assuming I have the following model in Django 1.8:

```python
import uuid
from django.db import models
from django.utils.translation import ugettext_lazy as _
from mptt.fields import TreeForeignKey
from mptt.models import MPTTModel

class MyModel(MPTTModel):
    id = models.UUIDField(default=uuid.uuid4, verbose_name=_('ID'))
    parent = TreeForeignKey(to='self', related_name='children', verbose_name=_('parent'))
    name = models.CharField(max_length=100, verbose_name=_('name'))

    class MPTTMeta:
        order_insertion_by = ['name']
```

As I am loading from a CSV File, and `TreeManager.move_node()` is called, I get the following error:
```text
  File "../env/lib/python2.7/site-packages/mptt/managers.py", line 1007, in _move_child_within_tree
    tree_id])
  File "../env/lib/python2.7/site-packages/debug_toolbar/panels/sql/tracking.py", line 174, in execute
    return self._record(self.cursor.execute, sql, params)
  File "../env/lib/python2.7/site-packages/debug_toolbar/panels/sql/tracking.py", line 104, in _record
    return method(sql, params)
  File "../env/lib/python2.7/site-packages/django/db/backends/utils.py", line 81, in execute
    return super(CursorDebugWrapper, self).execute(sql, params)
  File "../env/lib/python2.7/site-packages/django/db/backends/utils.py", line 65, in execute
    return self.cursor.execute(sql, params)
  File "../env/lib/python2.7/site-packages/django/db/utils.py", line 94, in __exit__
    six.reraise(dj_exc_type, dj_exc_value, traceback)
  File "../env/lib/python2.7/site-packages/django/db/backends/utils.py", line 65, in execute
    return self.cursor.execute(sql, params)
ProgrammingError: can't adapt type 'UUID'
```
The Reason
----------------
Because `node.pk` and `parent.pk` return a `uuid.UUID` instance in Python instead of the standard auto-incremented integer primary key (due to the model field calling [`to_python`](https://docs.djangoproject.com/en/1.7/ref/models/fields/#django.db.models.Field.to_python) on the value from the database):
```python
class TreeManager(models.Manager):
    ...
    def _move_child_within_tree(self, node, target, position)
        ...
        cursor = connection.cursor()
        cursor.execute(move_subtree_query, [
            left, right, level_change,
[...]

Issue #466, opened 2016-05-12
Unique constraint failure when moving nodes


olivierdalang:
Hi !

I have a MPTT tree with an unique constraint on the parent and a "code" field. With such a setup, I get an IntegrityError when I reparent a child if there's already a child with the code, **even if I change the code so that they don't conflict** !

This means I need to change the code, save, then reparent, and save again. This also means I'll get headaches if the new code is conflicting in the initial parent.

I'm surprised to find a bug with a setup that seems extremely common (wanting an unique property per parent). Did I miss something ?

Anyway, thanks a lot for the tool !

Here's the model I tested with :

``` python
class TestModel( MPTTModel ):
    class Meta:
        unique_together = (('parent','code',),)
    parent = TreeForeignKey('self', null=True)
    code = models.CharField(max_length=10)
```

And here's the steps to reproduce

``` python
from MyApp.models import TestModel
a = TestModel.objects.create(code='a',parent=None)
b = TestModel.objects.create(code='b',parent=None)
a1 = TestModel.objects.create(code='1',parent=a)
b1 = TestModel.objects.create(code='1',parent=b)

b1.code = '2'
b1.parent = a
b1.save() # THIS THROWS IntegrityError: UNIQUE constraint failed
```

olivierdalang:
I forgot to give the details :
mptt 0.8.4
django 1.9.6

olivierdalang:
And also note that in the admin, when having a Unique constrain violation, you get a nice error message (`Please correct the error below. Test model with this Parent and Code already exists.`), while here the exception isn't catched...

craigds:
Thanks for the report. mptt doesn't do anything special with unique constraints, and doesn't touch non-mptt attributes (like `code`) so I don't see how mptt could affect this really.

That said... to confirm that, could you please post the database queries involved, so we can tell what's going on? i.e. `connection.queries` (assuming `DEBUG` is on)

olivierdalang:
@craigds Here are the queries generated by the save() line (sqlite, but same occurs on postgres).
Clearly, mptt tries to update parent_id without updating the other fields (code...), leading to the constraint violation.

See my PR providing tests for this issue : https://github.com/django-mptt/django-mptt/pull/467

```sql

UPDATE "MyApp_testmodel"
SET "lft" = CASE
        WHEN "lft" > 3
            THEN "lft" + 2
        ELSE "lft" END,
    "rght" = CASE
        WHEN "rght" > 3
[...]

PR #468, opened 2016-05-17
Fix TreeManager._move_child_within_tree breaks with UUIDField() as primary key.

xyb:
Thanks for #357 and #322.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ dist
 MANIFEST
 tests/mydatabase
 *.egg
+.eggs
--- a/mptt/managers.py
+++ b/mptt/managers.py
@@ -73,6 +73,7 @@ def delegate_manager(method):
 
 
 class TreeManager(models.Manager.from_queryset(TreeQuerySet)):
+
     """
     A manager for working with trees of objects.
     """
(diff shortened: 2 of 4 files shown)


craigds:
Thanks. I've never used UUIDField myself, is the implication that `obj.pk` doesn't work for UUIDField for some reason? Do you know why that would be?

craigds:
I should read the linked things before replying. #357 describes the issue pretty well. It's interesting that django's UUIDField returns an object which doesn't adapt in psycopg2 😞

craigds:
Thanks for the patch and tests :)

I'm pondering where we should go with this. IMHO this is a less-than-ideal solution since it scatters the fix in several places in the codebase. It does work though :)

One alternative would be to rewrite the custom-SQL queries in django-mptt using the ORM. That wasn't possible back in the early days, but now we're supporting django 1.8+ we should be able to do it. Django 1.8 adds [Case/When](https://docs.djangoproject.com/en/1.9/ref/models/conditional-expressions/#conditional-expressions), more powerful F expressions, etc.

It looks like the methods that would need modifications are:

* `TreeManager._inter_tree_move_and_close_gap`
* `TreeManager._move_child_within_tree`
* `TreeManager._move_root_node`

It might make the huge queries even less readable, but it would solve this problem. I'm assuming of course that the ORM will be able to adapt its own UUID objects before passing them to the database backend. (Probably worth checking that before rewriting those huge SQL queries)

I'm leaning toward that solution since avoiding custom SQL will make the code more robust in general. Thoughts?

craigds:
Also see #466 which is an unrelated issue with the same query

xyb:
@craigds I agree with refactoring to using Django's ORM will be great for future developments, it's hard to accomplish all Django ORM's benefits with hand made SQL queries. But it should be done separately from this bug-fix. I have interest in the refactoring but do not have enough time recently.

matthiask:
@craigds I think we should merge this since this pull request simply makes django-mptt's code consistent. The improvement might still happen, but this pull request works and solves real problems right now.

michaeljohnbarr:
@xyb Great work! Thanks for taking this and running with it. @matthiask thanks for the merge!

matthiask:
Thanks @michaeljohnbarr , but you have to thank @craigds for the merge :)

PR #474, opened 2016-06-15
Add an 'all' argument to drilldown_tree_for_node.

dracos:
This adds a new optional argument (immediately after the required
arguments) to have the function return all descendants of the node,
not just its immediate children.

I added this because I wanted the cumulative count functionality of drilldown_tree_for_node, but wanted to show all the descendants of a particular node, not just its children. I'm not sure if there's a different/better way I could have done this, feel free to let me know :)

Rebased version of #246

### DIFF
--- a/docs/templates.rst
+++ b/docs/templates.rst
@@ -96,8 +96,8 @@ Populates a template variable with the drilldown tree for a given node,
 optionally counting the number of items associated with its children.
 
 A drilldown tree consists of a node's ancestors, itself and its
-immediate children. For example, a drilldown tree for a book category
-"Personal Finance" might look something like::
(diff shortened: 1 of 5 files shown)


ad-m:
It does not require the addition of the test?

dracos:
Sorry about that, I've added a test :)

matthiask:
Please treat this as a suggestion to be discussed: What about renaming the `all` argument to the template tag to `all_descendants` too? It adds verbosity but also clarity.

Apart from that: Looks good, thanks!

dracos:
Given that's what I called it in the code bits, I'd certainly be happy with that! Changed..

matthiask:
Nice, thanks!

Issue #132, opened 2011-06-20
insert_at does not support "using" kwarg


mvdwaeter:
I'd like to use the function insert_at with the kwarg using.
So that I can use another database setting.

TheDeadOne:
```diff
*** ./mptt/models.py	2016-03-04T18:09:24.734503+08:00
--- ./mptt/models.py	2016-03-14T20:15:18.604928+08:00
***************
*** 211,217 ****
                  # Fall back on tree id ordering if multiple root nodes have
                  # the same values.
                  order_by.append(opts.tree_id_attr)
!             queryset = node.__class__._tree_manager.filter(filters).order_by(*order_by)
              if node.pk:
                  queryset = queryset.exclude(pk=node.pk)
              try:
--- 211,218 ----
                  # Fall back on tree id ordering if multiple root nodes have
                  # the same values.
                  order_by.append(opts.tree_id_attr)
!             queryset = node.__class__._tree_manager.db_manager(node._state.db).filter(filters).order_by(*order_by)
              if node.pk:
                  queryset = queryset.exclude(pk=node.pk)
              try:
***************
*** 388,393 ****
--- 389,395 ----
      def __init__(self, *args, **kwargs):
          super(MPTTModel, self).__init__(*args, **kwargs)
          self._mptt_meta.update_mptt_cached_fields(self)
+         self._tree_manager = self._tree_manager.db_manager(self._state.db)
  
      def _mpttfield(self, fieldname):
          translated_fieldname = getattr(self._mptt_meta, fieldname + '_attr')
```

matthiask:
@TheDeadOne It would be awesome if you (or someone else) could submit a pull request with the relevant changes. Thanks in advance!

TheDeadOne:
@matthiask done #478

craigds:
thanks folks :)

matthiask:
Thanks!

PR #478, opened 2016-06-21
Multiple databases configuration support

TheDeadOne:
Fix for issue #132


jamalex:
Thanks so much for fixing this! We were facing the same issues.

@craigds Just wondering if you know when this is going to be included in a 0.8.5 release, so we can depend on the PyPI version in [Kolibri](https://github.com/learningequality/kolibri). Thanks!!

TheDeadOne:
@jamalex my pleasure :)

craigds:
@jamalex: 0.8.5 is now tagged and on PyPI 😄

jamalex:
Awesome, thanks so much!

PR #484, opened 2016-07-24
Reverting update in commit b0c8561876 due to an issue

Ameriks:
Issue is related with package django-modeltranslation.

If this line is active, then error is shown: AttributeError: Can't pickle local object 'add_manager.<locals>.patch_manager_class.<locals>.NewMultilingualManager'.

https://github.com/deschler/django-modeltranslation/blob/master/modeltranslation/translator.py#L188

I'm not sure - why is this line needed?

### DIFF
--- a/mptt/models.py
+++ b/mptt/models.py
@@ -394,7 +394,7 @@ class MPTTModel(six.with_metaclass(MPTTModelBase, models.Model)):
     def __init__(self, *args, **kwargs):
         super(MPTTModel, self).__init__(*args, **kwargs)
         self._mptt_meta.update_mptt_cached_fields(self)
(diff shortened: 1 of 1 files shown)


craigds:
Agreed that that line does actually look wrong - it's making a copy of the manager during `__init__` (i.e. per model instance), though managers are normally a class attribute. So yes I'm inclined to remove that line (not comment it out though)

cc @TheDeadOne, can you comment on why that line's there?

TheDeadOne:
I didn't find another way to support `using` method and `using` keyword (https://docs.djangoproject.com/en/dev/topics/db/multi-db/#manually-selecting-a-database).

PR #486, opened 2016-07-28
mk/nomagic branch as discussed

matthiask:
Lots of changes here. If reviewing gets too hard (or if you want to review smaller pieces) I'll happily split up the changes. 

(If that's not necessary I wont.)

### DIFF
--- a/mptt/admin.py
+++ b/mptt/admin.py
@@ -269,10 +269,7 @@ class DraggableMPTTAdmin(MPTTModelAdmin):
         all_nodes = {}
 
         mptt_opts = self.model._mptt_meta
-        items = queryset.values_list(
-            'pk',
-            '%s_id' % mptt_opts.parent_attr,
-        )
+        items = queryset.values_list('pk', 'parent_id')
         for p_id, parent_id in items:
             all_nodes.setdefault(
                 str(parent_id) if parent_id else 0,
--- a/mptt/admin.py
+++ b/mptt/admin.py
@@ -203,7 +203,6 @@ class DraggableMPTTAdmin(MPTTModelAdmin):
 
         return response
 
-    @transaction.atomic
     def _move_node(self, request):
         position = request.POST.get('position')
         if position not in ('last-child', 'left', 'right'):
@@ -223,13 +222,14 @@ class DraggableMPTTAdmin(MPTTModelAdmin):
             return http.HttpResponse('FAIL, no permission.')
 
         try:
-            self.model._default_manager.move_node(cut_item, pasted_on, position)
+            with transaction.atomic():
+                self.model._default_manager.move_node(cut_item, pasted_on, position)
         except InvalidMove as e:
             self.message_user(request, '%s' % e)
             return http.HttpResponse('FAIL, invalid move.')
         except IntegrityError as e:
             self.message_user(request, _('Database error: %s') % e)
-            raise
+            return http.HttpResponse('FAIL, invalid move.')
 
         self.message_user(
             request,
--- a/mptt/admin.py
+++ b/mptt/admin.py
@@ -75,13 +75,12 @@ class MPTTModelAdmin(ModelAdmin):
         # If this is True, the confirmation page has been displayed
         if request.POST.get('post'):
             n = 0
-            with queryset.model._default_manager.delay_mptt_updates():
-                for obj in queryset:
-                    if self.has_delete_permission(request, obj):
-                        obj.delete()
-                        n += 1
-                        obj_display = force_text(obj)
-                        self.log_deletion(request, obj, obj_display)
+            for obj in queryset:
+                if self.has_delete_permission(request, obj):
+                    obj.delete()
+                    n += 1
+                    obj_display = force_text(obj)
+                    self.log_deletion(request, obj, obj_display)
             self.message_user(
                 request,
                 _('Successfully deleted %(count)d items.') % {'count': n})
--- a/mptt/admin.py
+++ b/mptt/admin.py
@@ -75,7 +75,7 @@ class MPTTModelAdmin(ModelAdmin):
         # If this is True, the confirmation page has been displayed
         if request.POST.get('post'):
             n = 0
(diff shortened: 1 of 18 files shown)


craigds:
It occurred to me that `TreeManager.disable_mptt_updates` doesn't really add any magic; it just turns off recalculation of fields. Perhaps we could keep that even though we're dropping `delay_mptt_updates`

I'll ask the mailing list for further feedback on the changes. Thanks again

matthiask:
I've cleaned up the PR a bit, and removed the addition of bulk_load to keep the diff clean(er): (https://github.com/matthiask/django-mptt/tree/mk/nomagic-bulk-load)

matthiask:
While chasing down a surprising (at least in the beginning) crash I found a behavior change in the new code which must be documented, and I'm not yet clear on how to resolve all edge cases best.

The current code in django-mptt does NOT run `model.save()` within `TreeManager.move_node()`. The new code obviously has to run `model.save()` because the tree maintenance operations never touch the `parent` field, a consequence of the fix for https://github.com/django-mptt/django-mptt/issues/466.

Completely obvious in hindsight, but at first I was dumbfounded (we had crashes in [feincms3](https://github.com/matthiask/feincms3)'s pages administration and a [reproducible test failure](https://travis-ci.org/matthiask/feincms3/jobs/149103192) — more fixes forthcoming)

matthiask:
I have a few doubts regarding transaction isolation levels and reading MPTT values from the database. Not sure whether this approach is even viable. We should be alright with READ COMMITTED, but that's not the default on MySQL. 

(OTOH who even cares about correctness on MySQL. Sorry, could not resist.)

auvipy:
can this move forward?

mptt/admin.py
```diff
@@ -225,13 +234,14 @@ def _move_node(self, request):
...
-            self.model._tree_manager.move_node(cut_item, pasted_on, position)
+            with transaction.atomic():
+                self.model._default_manager.move_node(cut_item, pasted_on, position)
         except InvalidMove as e:
             self.message_user(request, '%s' % e)
             return http.HttpResponse('FAIL, invalid move.')
         except IntegrityError as e:
             self.message_user(request, _('Database error: %s') % e)
-            raise
+            return http.HttpResponse('FAIL, invalid move.')
```
orf: These should all have a 400 status code to go with them IMO.

mptt/managers.py
```diff
@@ -53,51 +52,40 @@
...
     """
     A manager for working with trees of objects.
     """
 
-    def contribute_to_class(self, model, name):
-        super(TreeManager, self).contribute_to_class(model, name)
+    @contextlib.contextmanager
+    def disable_mptt_updates(self):
+        # Preserve API
+        yield
```
orf: Maybe raise a deprecation warning here? It seems senseless to keep them around forever if they are defunct.

mptt/managers.py
```diff
@@ -794,46 +553,34 @@ def _inter_tree_move_and_close_gap(
         connection = self._get_connection(instance=node)
         qn = connection.ops.quote_name
 
-        opts = self.model._meta
         inter_tree_move_query = """
         UPDATE %(table)s
```
orf: I'm no expert on the internals of django-mptt, but why is raw sql needed here? Could you do:

```python
Model.objects.update(
    level=Case(When(lft__gte=x, then=F('level') - ....))),
    tree_id=Case(When(lft__gte=y...
)
```

Using the Case/When classes built into Django?

mptt/managers.py
```diff
@@ -952,62 +695,55 @@ def _make_sibling_of_root_node(self, node, target, position):
...
-                'table': qn(self.tree_model._meta.db_table),
-                'left': qn(opts.get_field(self.left_attr).column),
-                'right': qn(opts.get_field(self.right_attr).column),
-                'tree_id': qn(opts.get_field(self.tree_id_attr).column),
-            }
-            cursor = connection.cursor()
-            cursor.execute(space_query, [target, size, target, size, tree_id,
-                                         target, target])
+        space_query = """
+        UPDATE %(table)s
```
orf: Same for this?

mptt/managers.py
```diff
@@ -1123,41 +854,30 @@ def _move_child_within_tree(self, node, target, position):
...
-                WHEN %(right)s >= %%s AND %(right)s <= %%s
-                  THEN %(right)s + %%s
-                ELSE %(right)s END,
-            %(parent)s = CASE
-                WHEN %(pk)s = %%s
-                  THEN %%s
-                ELSE %(parent)s END
-        WHERE %(tree_id)s = %%s""" % {
+        SET level = CASE
+                WHEN lft >= %%s AND lft <= %%s
```
orf: And this?

mptt/managers.py
matthiask: The reason is simple: The code is a bit older than Django's support for Case/When :)

Case/When should work too.

mptt/managers.py
matthiask: Yes.

matthiask:
I'm now quite sure that this PR does not work with transaction isolation levels SERIALIZABLE and REPEATABLE READ. 

I read initial values, send a big UPDATE over the wire and read the same row again in the same transaction. We still had problems with MPTT attributes getting out of sync with the code in this PR. So maybe there are some good ideas in here, but I wouldn't recommend merging this as it is.

auvipy:
certainly this will need more works

orf:
Perhaps we should split it up into more manageable chunks? Less magic is certainly a good thing to strive for, there seems to be a fair bit here that can be made into separate MRs and reviewed/merged independently?

craigds:
Needs a bit of work put into bringing this up to date with master. Some commits are no longer relevant and others need a bit of refactoring.

I've started rebasing some of the commits here onto #567. That one will stay up-to-date with any master changes until it's merged.

I've started with the initial stable things that are obvious and I understand well, and I'll create some individual pull requests for the rest.

PR #491, opened 2016-08-03
Updated the "mptt_items_for_result" function

ataylor32:
The `mptt_items_for_result` function has diverged quite a bit from Django's `items_for_result` function. This change fixes that. My motivation for doing this was to get the `add_preserved_filters` functionality into django-mptt. This is a feature that Django has had since version 1.6, as noted towards the bottom of [the "Minor features" section of the Django 1.6 release notes](https://docs.djangoproject.com/en/dev/releases/1.6/#minor-features).


craigds:
Thanks!

PR #506, opened 2016-08-26
Fix #420

### LINKED ISSUE #420: Admin list_filter breaking `cache_tree_children`
Consider I have 3 objects based off MPTT with a boolean field named `active`.

The first object is 0 level object with `active` set to False.
The second object is a 1 level object with the parent being the first object and with `active` set to True.
The third object is a 0 level object with `active` set to True.

On the admin list page we now add a `list_filter` on the field `active` (i.e. `list_filter = ['active', ]`) and filter the status of the objects to have `active` being True.

This will return a queryset containing the second and third object.

When reaching `cache_tree_children` the queryset then gets ordered by the following line:

```
queryset = queryset.order_by(tree_id_attr, left_attr)
```

The queryset will now be ordered to have the second object first in the queryset then the third object.

The function then iterates the queryset setting the `root_level` as the node level of the first object in the queryset (in this case 1) and proceeds to loop.

The second object now has a node level of zero and hits the following conditional:

```
if node_level < root_level:
    # ``queryset`` was a list or other iterable (unable to order),
    # and was provided in an order other than depth-first
    raise ValueError(
        _('Node %s not in depth-first order') % (type(queryset),)
     )
```

This will now compare the current `node_level` 0 and the `root_level` 1 and raise the `ValueError`.

From my perspective once the list has been filtered the tree structure does not make semantic sense and should probably be returned as a normal list view containing the filtered objects as a suggestion.

Aramgutang:
see Issue [#420](https://github.com/django-mptt/django-mptt/issues/420) description and commit [2477e2ad8](https://github.com/ixc/django-mptt/commit/2477e2ad8b4f3b007a39d53eb17f863e5fb8f079) on master branch

### DIFF
--- a/mptt/templatetags/mptt_tags.py
+++ b/mptt/templatetags/mptt_tags.py
@@ -247,10 +247,14 @@ def cache_tree_children(queryset):
 
     Returns a list of top-level nodes. If a single tree was provided in its
     entirety, the list will of course consist of just the tree's root node.
+
+    For filtered querysets, if no ancestors for a node are included in the
+    queryset, it will appear in the returned list as a top-level node.
     """
 
     current_path = []
     top_nodes = []
+    is_filtered = False
 
     # If ``queryset`` is QuerySet-like, set ordering to depth-first
     if hasattr(queryset, 'order_by'):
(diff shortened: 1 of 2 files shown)


ad-m:
@Aramgutang , tests fails. Do you can fix that?

Aramgutang:
Yes, writing a slightly better solution now, will discuss in #420 when I'm done.

PR #518, opened 2016-10-21
FIX #517: TreeManager get_queryset_descendants and get_queryset_ancestors

### LINKED ISSUE #517: TreeManager.get_queryset_descendants() returns some of the nodes in the `queryset` parameter when `include_self`=False
This is currently the case in 0.8.6

It happens when there are more than two siblings provided in `queryset` **and** the siblings are on level=1.

I have provided a test case below to demonstrate this behavior. For the sake of brevity I assume you can drop this in some test app and just [disable migrations](https://docs.djangoproject.com/en/1.9/ref/settings/#migration-modules) or you may need to move the model out. This test will pass if you only make two nodes at level 1 or if you move the siblings to another level. I did not test it with "fuller" ancestor trees - only in the manner indicated.

```
from django.test import TestCase

from mptt import models


class Node(models.MPTTModel):
    parent = models.TreeForeignKey('self', null=True, blank=True,
                            related_name='children', db_index=True)


class MPTTModelTestCase(TestCase):
    def test_get_queryset_descendants_include_self(self):
        """Fails when siblings>=3 and sibling_level=1."""
        siblings = 3
        sibling_level = 1

        parent = Node.objects.create(level=0)
        for i in range(1, sibling_level):
            parent = Node.objects.create(level=i)
        for i in range(siblings):
            level_node = Node.objects.create(parent=parent)
            Node.objects.create(parent=level_node)

        level_qs = Node.objects.filter(level=sibling_level)
        manager_qs = list(Node.objects.get_queryset_descendants(
                level_qs, include_self=False))

        # For illustration
        manager_qs_include_self = list(Node.objects.get_queryset_descendants(
                level_qs, include_self=True))

        per_node = []
        for node in level_qs:
            per_node.extend(node.get_descendants(include_self=False))

        self.assertItemsEqual(manager_qs, per_node)
```

Some illustrative output on this failed case:

```
In [2]: [node.id for node in manager_qs]
Out[2]: [3L, 4L, 5L, 7L]

In [3]: [node.id for node in manager_qs_include_self]
Out[3]: [2L, 3L, 4L, 5L, 6L, 7L]

In [4]: [node.id for node in per_node]
Out[4]: [3L, 5L, 7L]
```

emilhdiaz:
Needed to take into account the lowest or highest level of a sibling when constructing the query for ancestors or descendants respectively. All unit tests are passing and added a new test as reported by  @thecardcheat

### DIFF
--- a/mptt/managers.py
+++ b/mptt/managers.py
@@ -148,11 +148,14 @@ class TreeManager(models.Manager.from_queryset(TreeQuerySet)):
         if direction == 'asc':
             max_attr = opts.left_attr
             min_attr = opts.right_attr
+            level_dir = max_op
         elif direction == 'desc':
             max_attr = opts.right_attr
             min_attr = opts.left_attr
+            level_dir = min_op
 
         tree_key = opts.tree_id_attr
+        level_key = '%s__%s' % (opts.level_attr, level_dir)
         min_key = '%s__%s' % (min_attr, min_op)
         max_key = '%s__%s' % (max_attr, max_op)
 
@@ -163,6 +166,7 @@ class TreeManager(models.Manager.from_queryset(TreeQuerySet)):
             min_attr,
             max_attr,
             opts.parent_attr,
+            opts.level_attr,
             # These fields are used by MPTTModel.update_mptt_cached_fields()
             *[f.lstrip('-') for f in opts.order_insertion_by]
         )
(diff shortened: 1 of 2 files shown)


craigds:
I'm not sure I follow what the issue is? The test seems flawed since it creates nodes with no parent but `level > 0`, which doesn't make much sense.

If you could fix the test to not do that, then I could run it against the 0.8.6 code to see the error. At present I can probably assume the bug is caused by the broken tree fields after creating a root node with level >0...

emilhdiaz:
@craigds good catch, I updated the test to construct a proper tree and I added more assertions to cover for when `include_self` is True or False.

emilhdiaz:
bump

musashiXXX:
I know I said I'd look at this, but I've been incredibly busy at work... I
simply haven't had the time, and honestly I don't know when I will.

emilhdiaz:
No worries, just wanted to make sure it didn't get lost! Take your time!

musashiXXX:
At this rate, it could be a few months before I get time :-(





------------------------------

*Charles Hamilton*
*musashi@nefaria.com <musashi@nefaria.com>*

musashiXXX:
@emilhdiaz This is still on my to-do list, fyi. I haven't forgotten about it, but I still don't have an ETA regarding when I'll be able to take a closer look. Time is a tricky thing, and perpetually in short supply.

dbramwell:
@emilhdiaz - King amongst men! Fixed my `get_queryset_ancestors` woes. Shame this hasn't been merged in

emilhdiaz:
@dbramwell I'm glad someone else got some use out of this! @musashiXXX is this MR still of interest? Should I update to resolve the merge conflicts?

dbramwell:
@emilhdiaz unfortunately it seems it's no longer actively maintained as of a month ago, so might not be worth your effort

musashiXXX:
@dbramwell @emilhdiaz I apologize for my lack of involvement the past few years, I _have_ been following this conversation though. 

>unfortunately it seems it's no longer actively maintained as of a month ago,

[This](https://406.ch/writing/django-tree-queries/) was quite an interesting read! While it's a sad to see such a long-lived project die, that's the nature of the work we do. 

So long, and thanks for all the fish!

cclauss:
This branch has git conflicts that must be resolved

matthiask:
Thank you for this contribution — this was a real and well-described bug with a correct fix! We independently implemented the same approach (adding a level constraint to each bounding-box range filter) and landed it directly on main. The fix and a regression test are now in the codebase. Sorry for not catching your PR at the time and crediting you properly — the issue tracker was a bit overwhelming.

PR #527, opened 2016-11-24
Finnish translation

TaijaQ:
I saw there was no Finnish translation, so I added support for the language. Anything that helps share this wonderful utility far and wide!


matthiask:
Thank you!

PR #528, opened 2016-12-19
Use ModelSignal over plain Signal

orf:
ModelSignal lets you specify the sender as a dotted string (i.e `app_label.ModelName`) which really helps with circular import weirdness.

Edit: It appears pypy builds are failing randomly. Any ideas?


craigds:
Thanks. Not sure about pypy builds. They've been doing that for a while though, not related to this change.

PR #533, opened 2017-01-05
dk to da

scootergrisen:


scootergrisen:
Please make sure the "dk" folder is deleted so only the "da" folder remains.

craigds:
Thanks for the PR

Can you link to something explaining why to use DA instead of DK? AFAIK these are normally labelled with ISO-3166-1 codes, but I can't seem to actually find a citation for that at the moment.

scootergrisen:
https://www.gnu.org/software/gettext/manual/gettext.html#Header-Entry

Reason why i wrote is because i found a few packages using it "dk":
https://www.debian.org/international/l10n/po/dk

Comare that will how many use "da" and "da_DK":
https://www.debian.org/international/l10n/po/da
https://www.debian.org/international/l10n/po/da_DK

"dk" is short for denmark. A country.
"da" is short for danish. A language.

If the translation is placed in /usr/share/locale/dk/LC_MESSAGES/ it is not used at all i believe.

claudep:
I can confirm the language code for Danish is `da`, not `dk`.

craigds:
Thanks. My mistake :)

PR #534, opened 2017-01-12
Use default tree manager when multiple tree managers are defined

### LINKED ISSUE #499: Models with multiple tree managers
While trying to upgrade to Django 1.9 and MPTT 0.8.5 from Django 1.8 and MPTT 0.7.0 I ran into some MPPT behaviour that I found very confusing.  Finally today I tracked down the issue.  We have a model with two TreeManager's on it like so:

```
class FooObjects(TreeManager):
    def get_queryset(self):
        return super(FooObjects, self).get_queryset().exclude(foo='')

class Bar(MPTTModel):
    parent = models.ForeignKey('self', related_name="children", ...)
    bar = CharField(...)
    foo = CharField(...)

    objects = TreeManager()
    foo_objects = FooObjects()

```

This worked fine in MPTT 0.7.0 but when upgrading to 0.8.5 weird things like this would happen:

```
bar1 = Bar.objects.create(bar="b1", foo='')
bar2 = Bar.objects.create(bar="b2", foo='')
bar3 = Bar.objects.create(bar="b3", foo='')

bar2.parent = bar1
bar2.save()
bar3.parent = bar1
bar3.save()

# this  should have created a parent (f1) with two children (f2, f3)

bar1.refresh_from_db()
bar2.refresh_from_db()
bar3.refresh_from_db()

bar2.parent # bar1
bar3.parent # bar1
bar1.get_children() # empty list !
bar1.get_ancestors() # empty list !
bar1.children.all() # [<bar2>, <bar3>] !!!

```

Eventually I figured out that the `_tree_manager` attribute of the `Bar` class was being set to `Bar.foo_objects` rather than `Bar.objects` as I was expecting.

`MPTTModelBase` [decides what to use as the _tree_manager_attribute](https://github.com/django-mptt/django-mptt/blob/master/mptt/models.py#L337-L360) based on the first instance of a `TreeManager` it finds in the list of `cls._meta.concrete_managers + cls._.meta.abstract_managers`.  Looking at `Bar._meta.concrete_managers` I found `[(35, 'foo_objects', <FooObjects>), (34, 'objects', <TreeManager>), (36, '_base_manager', <django.db.models.manager.Manager>)]` which explains why `_tree_manager` is set to `FooObjects` rather than `objects` as expected.

As a hacky workaround you can simply do the following:

```
class Bar(MPTTModel):
        ...

Bar._tree_manager = Bar.objects
```

or

```
class Bar(MPTTModel):
    ...
    __getattribute__(self, attr):
        if attr == "_tree_manager":
            return Bar.objects
        return super(Bar, self).__getattribute__(attr)
```

but neither of those is ideal from a future proofing perspective.

All of this to say, it would be great if at a minimum this was documented somewhere or even better if we could specify what the default tree manager should be.  Checking if `cls._default_manager` is an instance of `TreeManager` before falling back to `cls._meta.concrete_managers + cls._meta.abatract_managers` would probably work well too.

sephii:
Fixes #499. Bug only happens on Django < 1.10.

### DIFF
--- a/mptt/models.py
+++ b/mptt/models.py
@@ -338,18 +338,23 @@ class MPTTModelBase(ModelBase):
             if not abstract:
                 # make sure we have a tree manager somewhere
                 tree_manager = None
-                if hasattr(cls._meta, 'concrete_managers'):  # Django < 1.10
-                    cls_managers = cls._meta.concrete_managers + cls._meta.abstract_managers
-                    cls_managers = [r[2] for r in cls_managers]
+                # Use the default manager defined on the class if any
+                if cls._default_manager and isinstance(cls._default_manager, TreeManager):
+                    tree_manager = cls._default_manager
                 else:
-                    cls_managers = cls._meta.managers
+                    if hasattr(cls._meta, 'concrete_managers'):  # Django < 1.10
+                        # Django < 1.10 doesn't sort managers
+                        cls_managers = sorted(cls._meta.concrete_managers + cls._meta.abstract_managers)
+                        cls_managers = [r[2] for r in cls_managers]
+                    else:
+                        cls_managers = cls._meta.managers
 
-                for cls_manager in cls_managers:
-                    if isinstance(cls_manager, TreeManager):
-                        # prefer any locally defined manager (i.e. keep going if not local)
-                        if cls_manager.model is cls:
-                            tree_manager = cls_manager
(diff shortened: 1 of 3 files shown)


yakky:
Confirmed!

craigds:
thanks 😄

Issue #536, opened 2017-01-28
PR #491 broke raw_id_fields on Django 1.9 and below


yakky:
Changes in #491 broke the related selection from raw_id_fields by adding Django 1.10-only syntax (``data-popup-opener``).
The old ``onclick="opener.dismissRelatedLookupPopup(window, &#39;%s&#39;); return false;"`` syntax should be present for lower django versions

yakky:
I can provide a PR, I just need an advice how to do django-version detection according to the project code style

craigds:
This was fixed by #537, closing

PR #537, opened 2017-01-29
Fix dismissRelatedLookupPopup issue with Django 1.9 and below

yakky:
This is a preliminary fix for #536 
Tested to work on Django 1.8, 1.9 and 1.10

### DIFF
--- a/mptt/templatetags/mptt_admin.py
+++ b/mptt/templatetags/mptt_admin.py
@@ -182,12 +182,21 @@ def mptt_items_for_result(cl, result, form):
                 else:
                     attr = pk
                 value = result.serializable_value(attr)
+                if cl.is_popup:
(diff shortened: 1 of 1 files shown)


yakky:
@craigds could you check if it's an acceptable fix for the issue? If so, I'll check the tests and I will make this mergeable

craigds:
Thanks! Since you asked, I'd prefer a simple version check than the feature-checking check. Something like:

```python
if django.VERSION < (1, 10):
    opener = ...
```

yakky:
@craigds will comply and I will bring this in shape for merge. thanks for answering!

yakky:
@craigds fixed, thanks for the review

yakky:
Thanks!

PR #538, opened 2017-02-08
Fix missing django import

jaulin:
Found an issue with the missing django import. Please can you pull this in?


craigds:
Cheers (whoops!)

PR #539, opened 2017-02-11
Moved form-related tests in a separate file

claudep:


claudep:
In Django 1.11, the move to template-based form widgets will lead to small HTML layout differences which will be hard to account for in doctests. That's why I suggest to move to standard unit tests for all form tests.

craigds:
👍 those doctests are real pain, I'd get rid of them all if I had the time. Thanks for this!

PR #540, opened 2017-02-12
Test with Django 1.11

claudep:


.travis.yml
```diff
@@ -14,6 +14,7 @@ env:
   - DJANGO="Django>=1.8,<1.9"
   - DJANGO="Django>=1.9,<1.10"
   - DJANGO="Django>=1.10,<1.11"
+  - DJANGO="Django>=1.11a,<2.0"
```
felixxm: Small typo `1.11a` -> `1.11a1`

tox.ini
```diff
@@ -18,3 +18,4 @@ deps =
     dj18: Django>=1.8,<1.9
     dj19: Django>=1.9,<1.10
     dj110: Django>=1.10,<1.11
+    dj111: Django>=1.11a,<2.0
```
felixxm: Small typo `1.11a` -> `1.11a1`

.travis.yml
```diff
@@ -28,6 +29,12 @@ matrix:
       python: "3.2"
     - env: DJANGO="Django>=1.10,<1.11"
       python: "3.3"
+    - env: DJANGO="Django>=1.11a,<2.0"
+      python: "3.2"
+    - env: DJANGO="Django>=1.11a,<2.0"
+      python: "3.3"
+    - env: DJANGO="https://github.com/django/django/archive/master.tar.gz"
+      python: "2.7"
```
felixxm: Django 2.0 doesn't support `pypy`. Maybe you can add it to exclude matrix in this PR.

.travis.yml
claudep: Looks like it depends on the pip version… Thanks for the fix.

.travis.yml
claudep: I don't know enough about pypy to change anything related. I noticed though there is a pypy3 option to use on Travis.

craigds:
Yeah, I'm happy to remove `pypy` altogether. I doubt many people use it. In any case it behaves pretty much the same as CPython 2.7 from mptt's perspective, so no need to have it tested separately.

craigds:
but this change is good without that anyway. thanks!

.travis.yml
claudep: Despite the correction, it continues to fail (sometimes?).
https://travis-ci.org/django-mptt/django-mptt/jobs/203252723

craigds: super strange. Latest build worked. I'm hoping it stays that way. Possibly travis had cached something oddly. ¯\_(ツ)_/¯

PR #543, opened 2017-02-20
Fix print_debug_info with non-ascii object representation

sephii:
If you run `print_debug_info` and your objects have non-ascii characters in their unicode representation, you'll get a `UnicodeDecodeError`.

### DIFF
--- a/tests/myapp/tests.py
+++ b/tests/myapp/tests.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
 import io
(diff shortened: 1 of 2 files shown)


tests/myapp/tests.py
```diff
@@ -1573,14 +1574,27 @@ def test_alters_data(self):
...
 
+    def test_debug_info_with_non_ascii_representations(self):
+        Category.objects.create(name='El niño')
+
+        stream_type = self.get_stream_type()
+        with stream_type() as out:
+            print_debug_info(Category.objects.all(), file=out)
+            output = out.getvalue()
+
+        self.assertIn('El niño', output.decode('utf-8'))
```
craigds: this `.decode` is failing the tests on py3

sephii:
I had some trouble getting it to work properly on Python 2 and Python 3 because of the big changes about how encoding is handled in the csv module. According to http://python3porting.com/problems.html#csv-api-changes Python 2 expects a byte stream, while Python 3 expects a text stream. My new attempt reflects this, and works on both Python 2 and Python 3. There might be a cleaner way to do it, maybe through things like `TextIOWrapper` but I couldn't figure out a solution that would work on both Python 2 and Python 3 with the exact same code.

ad-m:
@sephii , see https://github.com/jdunck/python-unicodecsv . I hope it will be helpful.

craigds:
test failure is unrelated. thanks :)

PR #544, opened 2017-03-04
fix in template - <li> ... <li>

villian2008:
norm is
{% load mptt_tags %}
<ul class="main">
    {% recursetree nodes %}
        <li> {{ node.title }} </li>
            {% if not node.is_leaf_node %}
                <ul class="children">
                    {{ children }}
                </ul>
            {% endif %}
    {% endrecursetree %}
</ul>


matthiask:
Thanks!

According to https://www.w3.org/TR/html5/grouping-content.html#the-ul-element , only `li`, `script` and `template` are allowed as children of the `ul` element, which means that this change would promote invalid usage of HTML.

PR #546, opened 2017-03-04
Bumped Django version. Fixed old pip issue.

felixxm:


felixxm:
`pip>=8.0` doesn't support `python==3.2`. IMO we can drop python 3.2 support (see [[1]](https://docs.djangoproject.com/en/1.10/releases/1.8/#python-compatibility)).

craigds:
yes, can drop python 3.2 support. Honestly I don't think anyone has used python 3.2 for ages :)

craigds:
what's the old pip issue this fixes? Is that the bug causing travis to not find django 1.11a1 _sometimes_ ?

felixxm:
Yes, I think that `pip` upgrade should help.

craigds:
https://github.com/pypa/pip/issues/3511 suggests that pip 8 handles alpha versions _differently_ ¯\_(ツ)_/¯

I don't understand why the tests were passing _sometimes_. Do you know?

felixxm:
It's a mystery for me. I thought that it is a old<7 pip issue, but it's hard do say.

PR #549, opened 2017-03-06
Fixed tests on Django master.

felixxm:
I fixed tests on Django master, i.e.:
```python
File "django-mptt/django-mptt/tests/myapp/models.py", line 340, in UniqueTogetherModel
    parent = TreeForeignKey('self', null=True)
TypeError: __init__() missing 1 required positional argument: 'on_delete'
```


craigds:
👍  cheers

PR #552, opened 2017-03-25
Fetch CSRF token from hidden input field instead of cookie

zyv:
If Django parameter `CSRF_COOKIE_HTTPONLY` [1] is set as recommended by
`manage.py check --deploy`, client-side JavaScript will not to be able
to access the CSRF cookie [2].

This, in turn, breaks Ajax POST requests, so avoid trying to read it
from the cookie and fetch CSRF token from a hidden form field instead.

[1]: https://docs.djangoproject.com/en/1.10/ref/settings/#csrf-cookie-httponly
[2]: https://www.owasp.org/index.php/HttpOnly#What_is_HttpOnly.3F

Signed-off-by: Yury V. Zaytsev <yury@shurup.com>

### DIFF
--- a/mptt/static/mptt/draggable-admin.js
+++ b/mptt/static/mptt/draggable-admin.js
@@ -81,23 +81,6 @@ django.jQuery(function($){
         }
     }
 
-    /* Thanks, Django */
-    function getCookie(name) {
-        var cookieValue = null;
-        if (document.cookie && document.cookie != '') {
-            var cookies = document.cookie.split(';');
-            for (var i = 0; i < cookies.length; i++) {
-                var cookie = $.trim(cookies[i]);
-                // Does this cookie string begin with the name we want?
-                if (cookie.substring(0, name.length + 1) == (name + '=')) {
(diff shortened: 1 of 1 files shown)


zyv:
As far as I can see, the test failures are not caused by my changes. Also, for your information:

https://code.djangoproject.com/ticket/27534

Even if `CSRF_COOKIE_HTTPONLY` is off, apparently, getting the token from the hidden field will be also necessary if `CSRF_USE_SESSIONS` is on.

matthiask:
The test failures are probably fixed by this: https://github.com/django/django/commit/216bb2e8fbc3cb37615bbd70edaa73287acdca81

Anyway, I think the changes look good. I'm not 100% sure, but I think that as long as there are any actions on the model admin (not `actions = None`, https://docs.djangoproject.com/en/1.10/ref/contrib/admin/actions/#disabling-all-actions-for-a-particular-modeladmin) the proposed code should work just as good or even better as the current implementation.

craigds:
Sounds good to me. Thanks :)

zyv:
Thank you guys for maintaining `django-mptt`!

PR #557, opened 2017-04-03
Added option to DraggableMPTTAdmin to expand the tree on first page v…

### LINKED ISSUE #551: DraggableMPTTAdmin with expanded tree on startup
Is it possible to configure DraggableMPTTAdmin so that the tree is expanded on startup? At the moment it is collapsed and I can expand all nodes via the button (and then something is stored in the local storage of the browser) but I would like to change the user experience for my admin users here.

If it is currently not possible, you can treat this as a feature request (for this very nice project btw.). Thanks in advance.

normanjaeckel:
…isit. Closed #551.

### DIFF
--- a/mptt/static/mptt/draggable-admin.js
+++ b/mptt/static/mptt/draggable-admin.js
@@ -402,7 +402,9 @@ django.jQuery(function($){
                 expandOrCollapseNode(treeNode(storedNodes[i]));
             }
         } else {
-            collapseTree();
+            if (!DraggableMPTTAdmin.expandTreeByDefault) {
+                collapseTree();
+            }
         }
     }
 
--- a/docs/admin.rst
+++ b/docs/admin.rst
@@ -103,6 +103,11 @@ see below.)
 ``list_per_page`` is set to 2000 by default (which effectively disables
 pagination for most trees).
 
+You may set the attribute ``expand_tree_by_default = True`` in your
+DraggableMPTTAdmin to expand the tree on first page visit (default is
+False). After this the state of every node (expanded or collapsed) is saved
+to the browser storage.
+
 
 Replacing ``indented_title``
 ----------------------------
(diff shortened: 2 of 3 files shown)


mptt/static/mptt/draggable-admin.js
```diff
@@ -402,7 +402,11 @@ django.jQuery(function($){
                 expandOrCollapseNode(treeNode(storedNodes[i]));
             }
         } else {
-            collapseTree();
+            if (DraggableMPTTAdmin.expandTreeByDefault) {
+                expandTree();
```
matthiask: I don't think the expandTree call is necessary in this case since the tree should be fully expanded by default.

Maybe you could rename the feature to `collapse_tree_by_default = True` and simply do `DraggableMPTTAdmin.collapseTreeByDefault && collapseTree()` or something.

I don't care much either way though, I just think that since we have to do a thing for collapsing the tree and nothing for expanding it, using "collapse" to name the feature might be clearer.

normanjaeckel: You are right. I will remove the `expandTree()` clause.

The other point is that I thought you would not like to change the default behaviour of the class so that one have to decide to use this new feature to expand the tree on first visit. But then I recommend to use a flag that defaults to `False` so that one feels like opting in to a feature instead of opting out to a feature which is activated by default. I would always recommend to use flags that defaults to `False` in every programming context because doing this consequently reduces misunderstandings and silly bugs (a wise men thaught me once). Of course here it would not be a bug if we use `collapse_tree_by_default = True` instead of `expand_tree_by_default = False`. What about `dont_collapse_tree_on_first_visit = False`? (No problem if you don't like this ... :-))

So your project -> your choice ... ;-)

craigds: not a fan of `dont_collapse_tree_on_first_visit` - it's always better to minimise the number of negatives in var names to make them easier to understand.

Otherwise not concerned about collapse vs expand. This seems fine to me 👍

matthiask:
Thanks!

PR #561, opened 2017-04-22
Update doc links

tony:
Update docs to point to readthedocs.io

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -11,7 +11,7 @@ Django Models and working with trees of Model instances.
 
 Project home: http://github.com/django-mptt/django-mptt/
 
-Documentation: http://django-mptt.github.io/django-mptt/
+Documentation: https://django-mptt.readthedocs.io/
 
 Discussion group: http://groups.google.com/group/django-mptt-dev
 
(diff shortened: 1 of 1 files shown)


matthiask:
Thanks! I'm not sure about the change from github to readthedocs for the documentation -- I'll leave the final decision to @craigds.

craigds:
I never set up RTD. It looks like that RTD project is owned by @jbochi.

Since we'd need to potentially change the RTD config etc in future, I'd prefer not to change these unless we can get access to that.

@jbochi, how would you feel about surrendering the RTD project?

craigds:
i should note: gh-pages is a bit of a pain, so RTD would be better :)

tony:
I am not sure if gh-pages is updated by default, rtd is.

matthiask:
@craigds @jbochi Surrendering the project completely isn't necessary. You could also add additional maintainers to the project (https://readthedocs.org/profiles/craigds/ and https://readthedocs.org/profiles/mk/ would be nice)

tony:
Still think we should update to ReadTheDocs. It's automatically done on post-hook and other python docs have a home there

jbochi:
Sorry for the late reply. I've added @craigds and mk as maintainers. Should I add someone else?  I don't mind deleting the project if you want to recreate it too.

Issue #565, opened 2017-05-04
Add a composite index on tree_id and lft


orf:
As far as I can tell most `mptt` queries either filter or order on both `tree_id` and `lft` at once. It seems `mptt` adds `index=True` to all tree-related fields, however this is not too useful.

In our postgres-backed application adding a composite index like the one below improved query times tenfold, from ~150ms to ~10ms with 55638 rows and 2272 unique trees:

```sql
CREATE INDEX tree_idx
  ON public.tree
  USING btree
  (tree_id ASC, lft ASC);
```

With the new Django 1.11 index support adding this index in a cross-database way is as simple as adding `indexes = [models.Index(fields=['tree_id', 'lft'])]` in the `Meta` of the model.

Of course we could do this ourselves, but it seems like this would be a good thing to include as a default if possible. The individual indexes on `tree_id` and `lft` could be removed in this case as well.

craigds:
* this is great, and we should do this
* I need to get back to reviewing #486 and merging whatever bits of it we can for 0.9 (and planning the rest of it for 0.10 or 1.0).
* #486 will likely require migrations, as will this, so ideally I'd like to get them in the same release so that users don't have to do migrations twice.

I'm hoping I'll be able to have a closer look at #486 this weekend and add a bit more of a plan.

orf:
Awesome! I left a few comments on #486, just general things - feel free to resolve/close the discussions if they don't make sense in the context of this project.

I took a (failed) stab at implementing this in #566, I can try again after #486 is merged or stable enough to branch off.

craigds:
thanks!

orf:
Hey @craigds, I've added a merge request that adds support for this in #577, can you take a look?

craigds:
thanks @orf. sorry it took me so long to review that.

orf:
No problems! Thank you for meeting it.

PR #566, opened 2017-05-08
#565 - Add composite index to tree_id + lft

orf:
Not sure if this is the best way to do this, but this attempts to add a composite index on `tree_id` and `lft` as per #565

### DIFF
--- a/mptt/models.py
+++ b/mptt/models.py
@@ -3,6 +3,7 @@ from functools import reduce, wraps
 import operator
 import threading
 
+import django
 from django.db import models
 from django.db.models.base import ModelBase
 from django.db.models.query import Q
--- a/mptt/models.py
+++ b/mptt/models.py
@@ -341,6 +341,17 @@ class MPTTModelBase(ModelBase):
                     if field_name not in existing_field_names:
                         field = models.PositiveIntegerField(db_index=True, editable=False)
                         field.contribute_to_class(cls, field_name)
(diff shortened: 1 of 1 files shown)


mptt/models.py
```diff
@@ -341,6 +342,17 @@ def register(meta, cls, **kwargs):
                     if field_name not in existing_field_names:
                         field = models.PositiveIntegerField(db_index=True, editable=False)
                         field.contribute_to_class(cls, field_name)
+                        
+                if django.VERSION >= (1, 11):
+                    
+                    tree_id_attr = getattr(cls._mptt_meta, 'tree_id_attr')
```
jdufresne: I don't think you need `getattr()` here. `cls._mptt_meta.tree_id_attr` should do.

mptt/models.py
```diff
@@ -341,6 +342,17 @@ def register(meta, cls, **kwargs):
                     if field_name not in existing_field_names:
                         field = models.PositiveIntegerField(db_index=True, editable=False)
                         field.contribute_to_class(cls, field_name)
+                        
+                if django.VERSION >= (1, 11):
```
jdufresne: With recent changes, you can drop this guard. django-mptt now only suppors 1.11+.

matthiask:
This may be a good idea, but let's run with only the `tree_id` index for now. Thanks!

PR #575, opened 2017-06-14
442 bulk load

### LINKED ISSUE #442: Question / feature request: Bulk create tree from root node

cirotix:
As mentionned in #444 I have fixe some of @craigds requests. This is the corresponding PR.
Should you want a cleaner history I can rewrite it before the merge. I have provided atomic commits to make the changes clear, as the comments are not attached to this PR. 

[Resolves #442]


AMontagu:
Hi ! 

Why the function in this PR is not documented here  https://django-mptt.readthedocs.io/en/latest/mptt.managers.html ?

PR #577, opened 2017-06-15
Add 'tree_id', 'lft' to index_together

### LINKED ISSUE #565: Add a composite index on tree_id and lft

orf:
I hope this works! This actually closes #565.

Edit: oh god this turned out to be harder than expected. I'll give up on the github editor and just clone the repo tomorrow...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 sudo: false
 language: python
+cache: pip
+git:
+  depth: 1
 
 python:
   - "3.6"
--- a/mptt/models.py
+++ b/mptt/models.py
@@ -326,8 +326,10 @@ class MPTTModelBase(ModelBase):
 
                 bases.insert(0, MPTTModel)
                 cls.__bases__ = tuple(bases)
-
-            if _get_tree_model(cls) is cls:
+                
+            is_cls_tree_model = _get_tree_model(cls) is cls
+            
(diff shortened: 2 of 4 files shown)


orf:
@craigds I've got the tests to pass now. If you have some time could you take a look? This gives a *huge* performance boost to django-mptt applications. Like, really really really huge - we saw a 150% reduction in query time for almost all operations. Would be good to share and push a new release so everyone else can benefit!

orf:
ping @craigds, do you have any time to have a look at this? The changes are pretty small?

PR #589, opened 2017-08-22
Drop support for django 1.9

michael-k:
It reached its EOL in April 2017, see https://www.djangoproject.com/weblog/2017/apr/04/security-releases/


matthiask:
Thanks!

Not sure it is worth it, since you're not removing or simplifying any code here?

matthiask:
Closing this for now. We can revisit this later (not much later though) when considering dropping support for 1.8.

PR #592, opened 2017-09-14
Update overview.rst, added more links

ParvJain:
I've just added few links that were present in README.rst for the same context.


matthiask:
Thanks!

PR #603, opened 2017-10-23
Add Django to install_requires in setup.py

jdufresne:
As it is a runtime requirement, it should be in the list.


matthiask:
Thank you! It is a requirement, but it seems customary to not include the Django requirement in Django packages to make it easier to upgrade the package in a project without necessarily also updating Django itself (e.g. `pip install -U django-mptt`).

I'd like to keep it that way.

PR #604, opened 2017-10-23
Fix invalid escape sequences in string literals

jdufresne:
Starting with Python 3.6, invalid escape sequences in string literals are deprecated. When they exist, they produce a warning:

```
  DeprecationWarning: invalid escape sequence ...
```

For more information, see:

https://docs.python.org/3/whatsnew/3.6.html#deprecated-python-behavior

> A backslash-character pair that is not a valid escape sequence now generates a DeprecationWarning. Although this will eventually become a SyntaxError, that will not be for several Python releases.


matthiask:
Thanks!

PR #607, opened 2017-10-31
Have Travis CI use the tox configuration.

goodtune:
This reduces maintenance and allows contributors to test in the same way as the CI system. The `.travis.yml` is simpler as a result.

The tox configuration needed repair – it was failing for all virtual environments.

Removed the now unnecessary shell script & `requirements.txt` files by moving requirements to `setup.py` and `tox.ini` files.

This does however forgo testing against the `master` branch of Django.


matthiask:
Hi,

Thanks for your contribution!

I'm no fan of tox-travis. It hides the combination of versions with which tests fail, and only shows the Python version (which is most often less interesting when maintaining Django packages than the Django version)

Also, I really really like testing against the master branch as well (with `allow_failures`). 

Is there no way around this when using tox-travis?

jdufresne:
Just personal opinion, but I've found the Travis CI configuration style used by django-debug-toolbar to be preferable:

https://github.com/jazzband/django-debug-toolbar/blob/c7e79ade99d3c540896389ef86d1acf522c588f2/.travis.yml

Things I like:

- Uses same testing infrastructure as `tox.ini`, so local tests are the same as Travis tests
- There isn't a complex `exclude:` section in the configuration
- It is very explicit and easy to follow/modify
- It doesn't have any tox-travis magic

If others agree, I can open a PR that makes such a change.

PR #608, opened 2017-11-01
Added static path for custom media forms

bgomes:
I want to change static files to correct static path


matthiask:
Why is this necessary? The `static()` method is already applied to relative paths:

https://github.com/django/django/blob/stable/1.11.x/django/forms/widgets.py#L92

matthiask:
I'm closing this for now, but you can still provide a rationale for this change and reopen :)

Thanks!

PR #618, opened 2017-12-22
Update tutorial

pw94:
I updated tutorial file, because it was outdated.
The `context_instance parameter` in `render_to_response` was [deprecated in Django 1.8](https://docs.djangoproject.com/en/1.10/releases/1.8/#dictionary-and-context-instance-arguments-of-rendering-functions), and removed in Django 1.10.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -110,9 +110,7 @@ Make a view
 This one's pretty simple for now. Add this lightweight view to your ``views.py``::
 
     def show_genres(request):
-        return render_to_response("genres.html",
-                              {'nodes':Genre.objects.all()},
-                              context_instance=RequestContext(request))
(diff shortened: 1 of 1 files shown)


docs/tutorial.rst
```diff
@@ -127,7 +125,7 @@ Create a template called ``genres.html`` in your template directory and put this
 
     {% load mptt_tags %}
     <ul>
-        {% recursetree nodes %}
+        {% recursetree genres %}
```
matthiask: Renaming `nodes` to `genres` is fine, but you should also do this on the new line 113. Also, please add a single space after the `:` when creating the context dictionary while you're at it :)

Thanks!

pw94: Applied! :)

matthiask:
Thank you!

PR #619, opened 2017-12-26
Update draggable-admin.js

svandeneertwegh:
Added nice buttons instead of Listable links.


mikekeda:
"btn btn-info" - it will work only with bootstrap, no?

svandeneertwegh:
Yes, i tested it with django-suit django-mptt and django-reversion in one..

mikekeda:
so what the point to add those classes if it will not works without additional packages or css?

svandeneertwegh:
I suggested it because it looks ugly with the reversion based buttons and these are links wrapped in a LI in one row..  Django admin is bt3..

matthiask:
Thank you for your contribution! There are various problems with this pull request:

- The nl translation you added mostly has german strings, which isn't correct.
- The default Django admin templates wrap object actions with `<li>` tags. The removal of the `<li>` wrapping in your pull request breaks the HTML structure.

I might accept a pull request which makes the button addition a little bit smarter by maybe implementing variants for the default admin, grappelli and django-suit. However, the pull request in its current form introduces unacceptable regressions.

PR #622, opened 2018-01-20
Added MpttConfig

rsalmaso:
Added mptt.apps.MpttConfig.

### DIFF
--- a/mptt/__init__.py
+++ b/mptt/__init__.py
@@ -3,6 +3,8 @@ from __future__ import unicode_literals
 VERSION = (0, 9, 0)
 __version__ = '.'.join(str(v) for v in VERSION)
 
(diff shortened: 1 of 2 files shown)


matthiask:
Hmm, what for? We do not have any models and therefore no section in the Django admin panel so I don't immediately see how the appconfig class might be useful?

rsalmaso:
AppConfig is the _modern_ way to tell django that the python package is a django application, not only for models or admin panel.
There is no deprecation path from using `models.py` as a way to declare a django application, but django docs encourage to have an explicit AppConfig, and internally django create an AppConfig for you anyway.
Using the `default_app_config` there is no need to replace `mptt` to `mptt.apps.MpttConfig` in `INSTALLED_APP`, so backward compatibility is preserved.
I like to be explicit whenever they can.

PR #638, opened 2018-04-12
Document project as stable and ready for use in production

jdufresne:
Helps users know, at a glance, the state of the project. 

For a complete list of trove classifiers, see:

https://pypi.python.org/pypi?%3Aaction=list_classifiers


matthiask:
Thank you! I think that's fair to say. django-mptt is a tool with sharp edges, but that does not mean it's not production ready.

PR #648, opened 2018-04-19
[tests] Remove leftover code for Django < 1.11 support

michael-k:


tests/myapp/models.py
```diff
@@ -18,9 +18,6 @@ def custom_method(self):
 
 class CustomTreeManager(TreeManager):
 
-    def get_query_set(self):
-        return CustomTreeQueryset(model=self.model, using=self._db)
-
     def get_queryset(self):
         # Django 1.8 removed the fallbacks here.
```
jdufresne: This comment is now irrelevant. Can you remove it as well?

michael-k: Done; I also removed the method `get_empty_query_set` which is no longer needed.

matthiask:
@jdufresne Feel free to merge :)

Issue #651, opened 2018-05-10
django-mptt: bugs with multiple databases


DomHudson:
## Summary
Does django-mptt support multiple client databases? I am not expecting to be able to maintain trees between the databases but I am attempting to have each database own it's own independent tree structure.

The following code works as expected:
```python
database = 'default'

parent = Entity.objects.using(database).create()
child = Entity.objects.using(database).create()
child.parent = parent
child.save()
```

As does defining the entity with the parent during creation:
```python
database = 'default'

parent = Entity.objects.using(database).create()
child = Entity.objects.using(database).create(parent = parent)
```

--- 

However, as soon as the entity is not in the main database issues are encountered.
The first example now reads like this:
  ```python
database = 'anything-other-than-default'

parent = Entity.objects.using(database).create()
child = Entity.objects.using(database).create()
child.parent = parent
child.save()
```
This raises 
> mptt.exceptions.InvalidMove: A node may not be made a child of any of its descendants.

The second approach:
```python
database = 'anything-other-than-default'

parent = Entity.objects.using(database).create()
child = Entity.objects.using(database).create(parent = parent)
```
Raises the following:
```
  File "env/lib/python3.6/site-packages/django/db/models/query.py", line 394, in create
    obj.save(force_insert=True, using=self.db)
  File "env/lib/python3.6/site-packages/mptt/models.py", line 1015, in save
    self.insert_at(parent, position='last-child', allow_existing_pk=True)
  File "env/lib/python3.6/site-packages/mptt/models.py", line 738, in insert_at
    refresh_target=refresh_target)
  File "env/lib/python3.6/site-packages/mptt/managers.py", line 72, in wrapped
    return method(self, *args, **kwargs)
  File "env/lib/python3.6/site-packages/mptt/managers.py", line 556, in insert_node
    target._mptt_refresh()
  File "env/lib/python3.6/site-packages/mptt/models.py", line 1057, in _mptt_refresh
    )[0]
  File "env/lib/python3.6/site-packages/django/db/models/query.py", line 289, in __getitem__
    return list(qs)[0]
IndexError: list index out of range
```

If I go to line `env/lib/python3.6/site-packages/django/db/models/query.py` line 394 I can print out `self.db` and this points to `default` despite the all entities being created in a different database.

DomHudson:
If I replace [this line](https://github.com/django-mptt/django-mptt/blob/master/mptt/models.py#L1044) with the following it seems to fix the issue:

```python
[...]

PR #652, opened 2018-05-10
Bug fix: Multidatabase issues

DomHudson:
## Summary
This PR ensures that refreshing the MPTT model will retrieve the data from the database that is currently linked to the model.  Fixes https://github.com/django-mptt/django-mptt/issues/651

### DIFF
--- a/mptt/models.py
+++ b/mptt/models.py
@@ -1046,6 +1046,6 @@ class MPTTModel(six.with_metaclass(MPTTModelBase, models.Model)):
             opts.right_attr,
             opts.level_attr,
             opts.tree_id_attr,
-        ).first()
+        )[0]
         for k, v in values.items():
             setattr(self, k, v)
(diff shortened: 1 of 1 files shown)


matthiask:
What does .values().first() return when the queryset is empty? A dictionary, too?

The addition of using() is certainly correct.

DomHudson:
It returns `None` if the filter isn't met so it'll crash with 

```
AttributeError: 'NoneType' object has no attribute 'items'
```

Perhaps a change to simply using the `.get` method will work? Then it'll fail with a more appropriate `DoesNotExist` error. Or we can manually raise a DoesNotExist if the result `is None`. What do you think? Alternatively, I can just put the __getitem__ call back in. I'll make whichever change and update the PR.

DomHudson:
I've edited it to only add the `.using` method call for now.

matthiask:
Thank you!

DomHudson:
Hi, would it be a possibility to get this bug fix tagged - I have to keep the fix in my models until I can pull it from PyPi. 

Many thanks!
Dom

PR #675, opened 2018-09-26
Update upgrade.rst to include 0.9.1 changelog

adrienbrunet:
Hi, this changelog PR may be considered a work in progress. I haven't work on any of the changes and I just got the changes from the commit messages. Not sure if everything is relevant. I did my best but I'm pretty sure you could make it better. Hope that helps. Cheers !

### DIFF
--- a/docs/upgrade.rst
+++ b/docs/upgrade.rst
@@ -12,14 +12,14 @@ of Python and Django no longer receive security patches.
 0.9.1
 =====
 
-Support for django 2.1 has been added, support for django<1.11 is removed.
+Support for Django 2.1 has been added, support for Django<1.11 is removed.
 Support for deprecated South has been removed.
 
 Some updates have been made on the documentation such as:
 
 - Misc updates in the docs (formatting)
 - Added italian translation
-- Remove unnecessary db_index=True from doc examples
+- Remove unnecessary `db_index=True` from doc examples
 - Include on_delete in all TreeForeignKey examples in docs
 - Use https:// URLs throughout docs where available
 - Document project as stable and ready for use in production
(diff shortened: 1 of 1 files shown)


docs/upgrade.rst
```diff
@@ -9,6 +9,40 @@ UNRELEASED
...
+Misc:
+
+- Update pypi.python.org URL to pypi.org
+- Remove redundant tox.ini options that respecify defaults
+- Remove unused argument from _inter_tree_move_and_close_gap()
+- Trim trailing white space throughout the project
+- Pass python_requires argument to setuptools
+- Added MpttConfig
+- Add test case to support ancestor coersion callbacks. 
+- Extend tree_item_iterator with ancestor coersion callback.
```
matthiask: That should be "coercion"

docs/upgrade.rst
```diff
@@ -9,6 +9,40 @@ UNRELEASED
 End of life Python 3.3 and Django 1.8 are no longer supported. These versions
 of Python and Django no longer receive security patches.
 
+0.9.1
+=====
+
+Support for django 2.1 has been added, support for django<1.11 is removed.
```
matthiask: Django please (the Django project itself also always uses an uppercase D)

docs/upgrade.rst
```diff
@@ -9,6 +9,40 @@ UNRELEASED
...
+Support for deprecated South has been removed.
+
+Some updates have been made on the documentation such as:
+
+- Misc updates in the docs (formatting)
+- Added italian translation
+- Remove unnecessary db_index=True from doc examples
```
matthiask: Would be nice to use double backticks (``db_index=True``) so that code is formatted as such.

adrienbrunet:
It could indeed be much smaller. As I didn't know what was really relevant, I only took commit messages and tried to do my best with it. If you want to purge it to make it smaller (and maybe clearer), I don't mind at all. Thanks for reviewing this.

PR #680, opened 2018-12-05
check tree_model valid in contribute_to_class();

shouyong:
tree_model may be  None.

```python
AttributeError: 'NoneType' object has no attribute '_tree_manager'
```


matthiask:
LGTM

PR #690, opened 2019-02-16
Add Support for Django 2.1 and 2.2 as well as Python 3.7 and drop support for Python 3.4

michael-k:
Tests against Django master fail because Django 3.0 will no longer ship with `django.utils.six`.  The simplest solution would be to drop support for Python 2.7.  I'd happy to include this in this PR or a followup.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,17 @@
-sudo: false
+dist: xenial
 language: python
 cache: pip
 git:
   depth: 1
 
 python:
+  - "3.7"
   - "3.6"
   - "3.5"
   - "2.7"
 
 env:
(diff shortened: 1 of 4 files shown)


matthiask:
Thanks! I'm not completely sure what other improvements dropping Python 3.4 support enables, but I'm still fine with that :)

matthiask:
Also re. dropping 2.7 support: Adding `six` to `install_requires` would work too (and probably wouldn't lead to more requirements)

Things *are* made harder by the removals in `django.utils.*` though (and Django@master compatibility made me drop 2.7 in other projects too) so if it's either too hard or if you just want to drop 2.7 then feel free :)

PR #691, opened 2019-02-19
Drop support for Python 2

michael-k:


tests/myapp/tests.py
```diff
@@ -100,16 +97,7 @@ def flush(self):
...
-                    # unicode literals in the doctests screw up doctest on py3.
-                    # this is pretty icky, but I can't find any other
-                    # workarounds :(
-                    text = re.sub(r"""\bu(["\'])""", r"\1", text)
-                    temp.write(b(text))
-                else:
-                    temp.write(text)
-
+                temp.write(text.encode("latin-1"))
                 temp.flush()
```
matthiask: Is the temporary file even necessary at all now?

michael-k: No, it's not :)

docs/conf.py
```diff
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
```
jdufresne: You can drop these coding cookies as well. Python 3 defaults to decoding source files as utf-8.

craigds:
I haven't reviewed the code changes, but no objections to this in principle. It's time.

matthiask:
Thanks @michael-k!

goodtune:
Already merged but no objection either.

PR #697, opened 2019-04-13
fix media ordering

dantyan:
now for DraggableMPTTAdmin doesn't work js code
because `mptt/draggable-admin.js` loads before `admin/js/jquery.init.js`
and there is an error 

`Uncaught TypeError: Cannot read property 'fn' of undefined
    at draggable-admin.js:14`


matthiask:
Thanks!

PR #701, opened 2019-05-22
add_related_count: use Django Subquery, add extra_filters parameter

PetrDlouhy:
This code simplifies the `managers.py` code a lot by using Django Subqueries.
I have also added the `extra_filters` parameter, which enables filtering the counted items by specifying extra parameters like `extra_filters={'is_enabled': True}`

### DIFF
--- a/docs/models.rst
+++ b/docs/models.rst
@@ -380,7 +380,7 @@ Rebuilds the mptt fields for the entire table. This can be handy:
 It is recommended to rebuild the tree inside a ``transaction.atomic()`` block
 for safety and better performance.
 
-``add_related_count(queryset, rel_cls, rel_field, count_attr, cumulative=False)``
+``add_related_count(queryset, rel_cls, rel_field, count_attr, cumulative=False, extra_filters={})``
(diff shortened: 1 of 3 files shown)


PetrDlouhy:
@matthiask What do you think about this change? I think using subqueries is big leap forward for `django-mptt`.

matthiask:
I think this looks really good! Thank you.

It would be great if someone else could also review this though. I don't know the `Subquery` / `OuterRef` objects that well and worry a bit about breakage. The tests are all passing though so I'm already quite confident.

PetrDlouhy:
@matthiask I rebased to current master. Can I somehow raise your confidence :-)?

matthiask:
Let's trust the test suite :-) thank you!

schwiftyaf:
Do you have an idea when the next release will include this?

matthiask:
Unfortunately I do not have an idea. Each time I release a new package several people complain that the changelog is missing but nobody steps in to help.

This is a bit demotivating. It does not help either that I'm no heavy user of django-mptt anymore.

Maybe someone else finds the time and energy to do this?

schwiftyaf:
Thanks Matt. I'm an definitely an amateur here but will look into what this entails. I did test the updated managers.py file on my development site and it's working exactly as I had hoped.

PR #705, opened 2019-07-02
Allow column sort in MPTTModelAdmin

lorenzomorandini:
MPTT disables column sorting in django admin because it breaks the tree order. Still it would be nice to give the possibility to users to enable sorting if they want. In our case we needed to have sorting enabled so we made a few changes.

By default sorting is disabled but can be enalbed using
```mptt_change_list_sort = True```
in MPTTModelAdmin. A red message will appear above the change list results table, alerting the user that sorting by column will have consequences.

### DIFF
--- a/mptt/admin.py
+++ b/mptt/admin.py
@@ -41,6 +41,8 @@ class MPTTModelAdmin(ModelAdmin):
 
     form = MPTTAdminForm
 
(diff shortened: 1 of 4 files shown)


matthiask:
I don't enjoy the fact that this will force future mptt maintainers to reproduce changes made to the Django admin in mptt, so I'm going to reject this for now.

Issue #714, opened 2019-09-26
Logging tree change when using DraggableMPTTAdmin


duduklein:
When using DraggableMPTTAdmin, the changes are not logged using the LogEntry standard model that Django uses to log all admin actions.

Is it possible to log it? Would you accept a pull request?

matthiask:
Sure!

duduklein:
Done

PR #715, opened 2019-10-08
Logging move node action when using DraggableMPTTAdmin 

duduklein:
As mentioned in issue #714

I added the possibility to disable the logging, but left it ON by default, since it's django's admin defaut behavior.

There default change message has the format that django accepts to identify changed fields.

There are 2 methods to be overwritten if users want to customize the change message.

Let me know if it's OK for you

### DIFF
--- a/mptt/admin.py
+++ b/mptt/admin.py
@@ -3,18 +3,17 @@ import json
 from django import forms, http
 from django.conf import settings
 from django.contrib.admin.actions import delete_selected
-from django.contrib.admin.options import ModelAdmin
+from django.contrib.admin.options import ModelAdmin, IncorrectLookupParameters, get_content_type_for_model
+from django.contrib.admin.models import LogEntry, CHANGE
 from django.contrib import messages
 from django.db import IntegrityError, transaction
-from django.utils.encoding import force_text
+from django.utils.encoding import force_text, smart_text
 from django.utils.html import format_html, mark_safe
 from django.utils.translation import ugettext as _, ugettext_lazy
 from django.contrib.admin import RelatedFieldListFilter
 from django.contrib.admin.utils import get_model_from_relation
-from django.contrib.admin.options import IncorrectLookupParameters
 from django.core.exceptions import ValidationError
 from django.core.serializers.json import DjangoJSONEncoder
-from django.utils.encoding import smart_text
 from django.utils.translation import get_language_bidi
 from django.db.models.fields.related import ForeignObjectRel, ManyToManyField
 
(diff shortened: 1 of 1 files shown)


mptt/admin.py
```diff
@@ -113,6 +115,7 @@ class DraggableMPTTAdmin(MPTTModelAdmin):
     list_display_links = ('indented_title',)  # Sane defaults.
     mptt_level_indent = 20
     expand_tree_by_default = False
+    log_move_node_action = True
```
matthiask: Is there a reason for making this optional?

mptt/admin.py
```diff
@@ -4,6 +4,8 @@
 from django.conf import settings
 from django.contrib.admin.actions import delete_selected
 from django.contrib.admin.options import ModelAdmin
+from django.contrib.admin.options import get_content_type_for_model
```
matthiask: Lines 6 and 7 import from the same module.

mptt/admin.py
```diff
@@ -172,6 +175,13 @@ def changelist_view(self, request, *args, **kwargs):
 
         return response
 
+    def get_data_before_update(self, request, cut_item, pasted_on):
+        return {k: getattr(cut_item, k) for k in ("lft", "mptt_level", "rght", "tree_id")}
```
matthiask: The fields' names aren't necessarily `lft`, `rght` etc. -- they can be overridden unfortunately.

You should inspect `cut_item._mptt_meta`. Also, `mptt_level` isn't even the default name.

mptt/admin.py
duduklein: Backward compatibility only. If you think it's not necessary, I'll remove it. let me know

mptt/admin.py
duduklein: I thought you wanted like this, since there were already 2 imports from 2 modules in separate lines :
`from django.contrib.admin.options import ModelAdmin` (line 6) and 
`from django.contrib.admin.options import IncorrectLookupParameters` (line 16) 
and
`from django.utils.encoding import force_text` (line 11) and 
`from django.utils.encoding import smart_text` (line 19).

Do you want me to put all 5 imports (lines 6, 7 and 16) in 2 different lines (since we have 3 imports from one module and 2 from another)?

mptt/admin.py
duduklein: Really sorry, I completely missed this. We'll correct it and re-submit it.

mptt/admin.py
matthiask: My opinion is that it's better to avoid branches if possible -- if someone later objects they can still introduce the conditional.

mptt/admin.py
matthiask: Ah right, I didn't see that; in my own projects I'm using black and isort so that I don't even have to think about formatting and sorting imports. 

I prefer one import per module, but you don't have to fix the code around it. I generally prefer one change per pull request but given that django-mptt does not see many changes these days the chance of conflicts is really small, so feel free to fix the surrounding code as well :)

mptt/admin.py
matthiask: Don't worry, that's exactly what unittests and reviews are for :) 

Thanks!

mptt/admin.py
duduklein: Just pushed all the corrections. Let me know if something else is missing.

duduklein:
All tests have passed.

mptt/admin.py
```diff
@@ -172,6 +171,16 @@ def changelist_view(self, request, *args, **kwargs):
 
         return response
 
+    def get_data_before_update(self, request, cut_item, pasted_on):
+        mptt_opts = self.model._mptt_meta
+        mptt_attr_fields = ("parent_attr", "left_attr", "right_attr", "tree_id_attr", "level_attr")
+        mptt_fields = [getattr(mptt_opts, attr) for attr in mptt_attr_fields]
+        return {k: getattr(cut_item, k) for k in mptt_fields}
+
+    def get_mode_node_change_message_log(self, request, cut_item, pasted_on, data_before_update):
```
matthiask: Shouldn't this be `get_move_node_change_message`? move, not mode; also, `_log` seems unnecessary.

Looks good apart from that!

duduklein: I missed that typo. thanks!

matthiask:
Awesome, thanks!

PR #720, opened 2019-12-10
Removing deprecated functions for Django 3

theunraveler:
Since support for Python 2 was dropped (https://github.com/django-mptt/django-mptt/pull/691) and Django 3 has deprecated the various legacy unicode string functions (https://docs.djangoproject.com/en/3.0/releases/3.0/#features-deprecated-in-3-0), I just replaced the following:

* `ugettext_` -> `gettext_`
* `force_text` -> `force_str`
* `smart_text` -> `smart_str`


matthiask:
Thanks!

Issue #721, opened 2019-12-20
Add compatibility with Django 3.0


claudep:
Currently, mptt tests are failing with Django 3.0, apparently due to this Django commit:
https://github.com/django/django/commit/4122d9d3f198

hlth:
Also breaking change is the removal of django utils.six and other python 2 compatibility APIs

https://docs.djangoproject.com/en/3.0/releases/3.0/#removed-private-python-2-compatibility-apis

pallavg:
Could this issue be prioritized as it prevents django-wiki from working in a django 3.0 project? Thank you.

matthiask:
I don't know whether anybody is working on this currently. I'd certainly appreciate a pull request @pallavg

pallavg:
Hi @matthiask,

I had hacked the source code in the following files to get django-wiki to work on my Django 3.0.1 project. The main issue was django.utils.six ModuleError.

```
models.py:
#from django.utils import six
import six
```

```
utils.py:
#from django.utils.six import next, PY3, text_type
#from django.utils.six.moves import zip
from six import next, PY3, text_type
from six.moves import zip
```

I looked at the master branch and it seems like these imports have been totally removed. So I am not sure if there is anything further to be done as far django.utils.six is concerned.

claudep:
No, the main issue is not related to `six` issues, it is related to the fact that in Django 3.0, setting a value to an `instance.foreign_id` field is automatically setting the `instance.foreign` value. django-mptt was using the capability to have temporary non-matching `instance.foreign_id`/`instance.foreign`, which is not possible any more.

slinkymanbyday:
apologies for the spam, but is there a chance of a new release as this has been merged in?

matthiask:
Closed in #724

PR #724, opened 2020-01-09
Add Django 3 support

reupen:
#721

Following django/django@4122d9d, this changes the model `save()` logic to get `self.parent` before (rather than after) it temporarily sets `self.parent_id` to the old parent ID.

I'm not particularly confident of the fix, since I'm not sure how it was guaranteed that `self.parent` and `self.parent_id` are consistent prior to django/django@4122d9d.

Nonetheless, this gets tests passing, so if there are problems with this approach I think additional tests should be added.

(Additionally, I updated the test matrix for Python 3.8: https://docs.djangoproject.com/en/3.0/faq/install/#what-python-version-can-i-use-with-django.)

### DIFF
--- a/mptt/models.py
+++ b/mptt/models.py
@@ -910,15 +910,14 @@ class MPTTModel(models.Model, metaclass=MPTTModelBase):
                 same_order = True
 
             if not same_order:
+                parent = getattr(self, opts.parent_attr)
                 opts.set_raw_field_value(self, opts.parent_attr, old_parent_id)
                 try:
                     right_sibling = None
                     if opts.order_insertion_by:
-                        right_sibling = opts.get_ordered_insertion_target(
-                            self, getattr(self, opts.parent_attr))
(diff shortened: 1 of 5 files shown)


mptt/models.py
```diff
@@ -910,15 +910,14 @@ def save(self, *args, **kwargs):
                 same_order = True
 
             if not same_order:
+                parent = getattr(self, opts.parent_attr)
```
reupen: I assume this could cause a query, hence skipping it if `opts.order_insertion_by or parent_id is not None` is falsey might be an optimisation, but I'm not sure it's worth it.

reupen:
Just spotted that I missed adding the Python 3.8 classifier. Happy to squash that into the first commit where it belongs.

ciur:
Both changes do the trick for me.

OskarPersson:
Are there anything more blocking this from being merged (and a new release containing it)?

matthiask:
I think it's fine with the additional testing this PR received.

Thanks all!

PR #730, opened 2020-03-03
Fix tree caching with multiple roots

### LINKED ISSUE #658: get_cached_trees - Utility Function appears to not cache parents correctly if passed multiple trees

artscoop:
Took into account the `tree_id` attribute for an object since it is mandatory to in which tree the parent is.
Fixes #658


artscoop:
Well wait... It's still incorrect...

PR #731, opened 2020-03-04
Fix caching utility

### LINKED ISSUE #658: get_cached_trees - Utility Function appears to not cache parents correctly if passed multiple trees

artscoop:
Fixed `mptt.utils.get_cached_trees` returning incorrect data,
moreso when there is more than one tree. The function is more memory intensive,
but works in every case with the proper results in cache and with similar
performance.
Updated docstrings.
Python 3.5+ only.
Fixes #658

### DIFF
--- a/mptt/utils.py
+++ b/mptt/utils.py
@@ -6,11 +6,17 @@ import copy
 import csv
 import itertools
 import sys
+from typing import List, Optional, Tuple
 
+from django.db.models.query import QuerySet
 from django.utils.translation import gettext as _
 
-__all__ = ('previous_current_next', 'tree_item_iterator',
-           'drilldown_tree_for_node', 'get_cached_trees',)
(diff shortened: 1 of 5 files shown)


mptt/utils.py
```diff
@@ -6,11 +6,17 @@
 import csv
 import itertools
 import sys
+from typing import List, Optional, Tuple
```
matthiask: These seem to be unused? Typing may be a good idea but this should happen in a different PR/issue.

mptt/utils.py
```diff
@@ -6,11 +6,17 @@
...
 
+from django.db.models.query import QuerySet
```
matthiask: ... this seems unused too.

matthiask:
Replaced by #830

PR #733, opened 2020-03-08
docs: Fix simple typo, neccessary -> necessary

timgates42:
There is a small typo in docs/models.rst, mptt/managers.py, mptt/models.py.

Should read `necessary` rather than `neccessary`.


matthiask:
Thank you!

PR #738, opened 2020-04-15
Removed unused import

nautics889:
I couldn't fully understand if there is a real reason to be 
```from django.conf import settings```
at fifth line? I just was browsing sources of my project and accidentally encounter on this module and my editor introspector didn't show me any usage of this across the file.
Sorry if i'm mistaken.


matthiask:
Maybe `settings` was used earlier, it certainly isn't used now.

PR #743, opened 2020-05-14
Allow to use TreeQuerySet as manager method to get TreeManager class

ncopiy:
This pull request allows users to write something like:

```
class MyModel(MPTTModel):
    objects = MyTreeQuerySet.as_manager()
```

instead of:
```
class MyModel(MPTTModel):
    objects = TreeManager.from_queryset(MyTreeQuerySet)()
```

`objects` will be instance of TreeManager instead of standard django manager.  
It means, users will be able to use TreeManager methods in `objects`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ MANIFEST
 tests/mydatabase
 *.egg
 .eggs
+venv/
+.idea/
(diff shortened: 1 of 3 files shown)


matthiask:
Thanks!

PR #749, opened 2020-06-22
Fix local variable referenced before assignment

MahmoudFarid:
in case enabling `_mptt_use_cached_ancestors` and the node is a root, it will raise an error on this `getattr(p, opts.parent_attr)`


matthiask:
Thanks!

Issue #698, opened 2019-05-01
0.10.0 release notes missing


moggers87:
There's nothing in https://github.com/django-mptt/django-mptt/releases or https://github.com/django-mptt/django-mptt/blob/master/docs/upgrade.rst about that latest release.

DmytroLitvinov:
Also, please update "Latest release" to 0.10

sbor23:
There's also no "Upgrade Notes" entry on [readthedocs](https://django-mptt.readthedocs.io/en/latest/upgrade.html), even tho the upgrade requires migrations, in our case.

dwasyl:
This would definitely be helpful

Mogost:
django-mptt version 0.11.0 missed release notes again.

matthiask:
Yes, who wants to help out?

benjaoming:
I can see that `db_index=True` has been removed on a number of fields since 0.10.0. This causes migrations to be necessary on all applications with MPTT models and naturally should make people wonder what that migration is for.

The reason is here: https://github.com/django-mptt/django-mptt/pull/578

matthiask:
I added a changelog but the contents are still missing. Closing this issue.
https://github.com/django-mptt/django-mptt/blob/master/CHANGELOG.rst

PR #751, opened 2020-07-22
Remove providing_args argument is signals

ad-m:
> django.utils.deprecation.RemovedInDjango40Warning: The providing_args argument is deprecated. As it is purely documentational, it has no replacement. If you rely on this argument as documentation, you can move the text to a code comment or docstring.

### DIFF
--- a/mptt/signals.py
+++ b/mptt/signals.py
@@ -5,8 +5,4 @@ from django.db.models.signals import ModelSignal
 # ``move_to``.
 # If the signal is called from ``save`` it'll not be pass position.
 
-node_moved = ModelSignal(providing_args=[
-    'instance',
-    'target',
-    'position'
-])
+node_moved = ModelSignal()


matthiask:
Thanks!

Mogost:
@matthiask It looks like this is not yet included in the release. Could you please release a new version?

matthiask:
@Mogost Thanks for the reminder. Yes, I'd like to do that. 

Please don't take it personally because it isn't meant that way. The idea of getting even more complaints about missing release notes makes it really hard to just put out another release. See e.g. https://github.com/django-mptt/django-mptt/issues/698, https://github.com/django-mptt/django-mptt/issues/775

matthiask:
So, I pushed another release without release notes and at the same time proposed a pull request marking this project as unmaintained. See https://github.com/django-mptt/django-mptt/pull/776

PR #756, opened 2020-08-31
Drop a bunch of python2-specific code

### LINKED ISSUE #725: Some small nitpicks for python 3
Some improvements for #691
This option no need for py3 only packages
https://github.com/django-mptt/django-mptt/blob/bc4d80b966eca2e614b29de6be2ed1091b1bea01/setup.cfg#L13

That can be rewritten without additional parameters in py3 -> `return super().formfield(**kwargs) `
https://github.com/django-mptt/django-mptt/blob/8ef52b348c285b660b2b08e0a1c707b01e2c9f76/mptt/fields.py#L26

`u''` -> `''`
https://github.com/django-mptt/django-mptt/blob/bc4d80b966eca2e614b29de6be2ed1091b1bea01/mptt/utils.py#L65

-> `__str__`
https://github.com/django-mptt/django-mptt/blob/96df84079f65f8529134ae7f14c3f88b588811aa/tests/.coveragerc#L9

Out of scope but easy to fix: _(many places)_
https://github.com/django-mptt/django-mptt/blob/bc4d80b966eca2e614b29de6be2ed1091b1bea01/mptt/utils.py#L80
Do not call getattr(x, 'attr'), instead use normal property access: x.attr. Missing a default to getattr will cause an AttributeError to be raised for non-existent properties. There is no additional safety in using getattr if you know the attribute name ahead of time.

WhyNotHugo:
Fixes  #725

### DIFF
--- a/docs/forms.rst
+++ b/docs/forms.rst
@@ -51,7 +51,7 @@ The text used to indicate a tree level can by customised by providing a
 ``level_indicator`` argument::
 
    category = TreeNodeChoiceField(queryset=Category.objects.all(),
-                                  level_indicator=u'+--')
+                                  level_indicator='+--')
 
 ...which for this example would result in a select with the following
 options::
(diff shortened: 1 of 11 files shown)


matthiask:
Thanks!

PR #758, opened 2020-09-07
Use_state.adding in raise_if_unsaved

jlmadurga:
In order to check if an instance is unsaved in the decorator `raise_if_unsaved` we should use `_state.adding`. Current `not self.pk 
` is raising problems for custom primary keys.

https://docs.djangoproject.com/en/3.1/ref/models/instances/

PD: don't know if using `_is_saved` is a better idea to handle this logic in only one place


matthiask:
Thanks, makes sense!

PR #761, opened 2020-10-13
Simplify tox.ini

jaap3:
Simplify tox.ini, test against Django master with tox as well and reuse tox in travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-dist: xenial
 language: python
 cache: pip
 git:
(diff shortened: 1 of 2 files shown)


matthiask:
Hi @jaap3 

Thank you for your contribution!

I don't like tox-travis very much because when using it, the build jobs list is strictly worse than without it. 
https://travis-ci.org/github/django-mptt/django-mptt/builds/735406794 only shows Python versions, https://travis-ci.org/github/django-mptt/django-mptt/builds/724939226 shows Python AND Django versions.

Personally, I don't find it enormously useful to run all tests in all version combinations locally. I'm happy with running only "extreme" combinations locally and let Travis CI do the rest, e.g. https://github.com/matthiask/django-js-asset/blob/master/tox.ini

What do you think?

jaap3:
I agree, the output on Travis is less useful. The reason I did it this way is that most projects I've looked at try to keep Tox and Travis in sync. I'll close this PR and might open a new one with the approach you suggested if I have the time.

PR #762, opened 2020-10-15
Set pre version, to make installing git version over latest releae possible

ErwinJunge:
With the current state of master, when first installing via `pip install django-mptt` and then trying to upgrade to git master by installing via `pip install 'git+https://github.com/django-mptt/django-mptt.git'` pip thinks the version is the same (0.11.0) and therefore doesn't actually install the newer code.

This PR fixes that by making master a `-pre` version. The idea is that when releasing 0.12.0, the `-pre` is removed in the release branch and master can then be bumped to `0.13.0-pre`.

### DIFF
--- a/mptt/__init__.py
+++ b/mptt/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.11.0"
+__version__ = "0.12.0-pre"
 VERSION = tuple(__version__.split("."))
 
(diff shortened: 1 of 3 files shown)


matthiask:
Yeah, I'm going to reject this for now. I don't want to do the additional work (even though it's not much). If someone else steps up this may yet happen. See https://github.com/django-mptt/django-mptt/issues/833

PR #776, opened 2021-03-04
Mark this project as unmaintained

matthiask:
See https://twitter.com/matthispam/status/1367205308439949313 and all the reports about missing release notes etc.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,6 +2,15 @@
 **This project is currently unmaintained**
 ==========================================
 
+Alternatives to django-mptt include:
+
+* `django-treebeard <https://pypi.org/project/django-treebeard/>`_ includes a MPTT
+  implementation (called nested set)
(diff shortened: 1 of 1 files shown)


claudep:
I think suggesting replacement packages should be very nice!

matthiask:
> I think suggesting replacement packages should be very nice!

Thanks, that's an excellent suggestion!

claudep:
Awesome, thanks! And many thanks also for all the energy you put in this package until now. It is appreciated!!

craigds (approved):
Great call. I think the alternative projects do a better job these days anyway. I came to the realisation at some point that MPTT can't really be managed correctly in the app layer, doing it in the db or just using CTEs seems like the way to go.

Thanks so much for putting in all the work to maintain this project. Hope 2021 improves for you

goodtune (approved):

PR #781, opened 2021-03-17
Update the changelog

rtpg:
This might be a bit incomplete, but I did my best to highlight what
 would be relevant to package users based on looking over the commit
 log.

 It's pretty nice to have the readthedocs-facing stuff, as well as the
 the `CHANGELOG.rst`, so I just copy-pasted the same content. I'm sure
 some people would do this in a bit of a cleaner way.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,21 +7,31 @@ Next version
 0.12
 ~~~~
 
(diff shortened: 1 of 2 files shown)


matthiask:
Thanks, this is great!

PR #787, opened 2021-06-16
Always call get_ordered_insertion_target to get right sibling

claudep:
This allows custom subclasses to implement a custom logic to
get insertion target.


matthiask (approved):
LGTM.

PR #795, opened 2021-08-20
Fix to return only concrete fields via _get_user_field_names

### LINKED ISSUE #794: HTTPModel._get_user_field_names() returns non-concrete fields returning in ValueError on save in Django
When you have a field which is non-concrete (practically meaning it has no name defined, which is a trick you can use when working with proxys on the field, like in https://github.com/bikeshedder/django-composite-field), django-mptt will return it through the internal function `_get_user_field_names()` and then pass it to the `save()` Method of Django, resulting in a `ValueError`. See https://github.com/django/django/blob/main/django/db/models/base.py#L703-L717

MRigal:
Fixes #794

### DIFF
--- a/mptt/models.py
+++ b/mptt/models.py
@@ -865,7 +865,7 @@ class MPTTModel(models.Model, metaclass=MPTTModelBase):
             self._mptt_meta.tree_id_attr,
             self._mptt_meta.level_attr,
         )
-        for field in self._meta.fields:
+        for field in self._meta.concrete_fields:
             if (
                 (field.name not in internal_fields)
                 and (not isinstance(field, AutoField))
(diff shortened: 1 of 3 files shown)


matthiask (approved):
Looks good to me. At the same time I'm left wondering, why does the code of django-mptt have to be this complex 😅

MRigal:
Why are so many Django Libraries so complex 😆 ... But yes, there wouldn't be an issue if it was not trying to reimplement so many Django methods.

If you feel like writing an article on why (I got it) and how to shift from django-mptt to django-tree-queries, you'll have a very interested reader ;-)

matthiask:
Here's an issue with an outline but it's definitely not an article yet... :) 
https://github.com/matthiask/django-tree-queries/issues/3

PR #796, opened 2021-08-21
Do not install tests.myapp

mgorny:
Fix options.packages.find.exclude to cover subpackages of "tests"
recursively.  Otherwise, tests.myapp is installed.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -44,7 +44,7 @@ tests =
 [options.packages.find]
 exclude =
     tests
-    myapp
+    tests.*
 
 [flake8]
 exclude = venv,.tox,build,docs


setup.cfg
```diff
@@ -44,6 +44,7 @@ tests =
 [options.packages.find]
 exclude =
     tests
+    tests.*
     myapp
```
matthiask, line 48:
```suggestion
```

This can probably be removed now?

matthiask (approved):
Oh wow, excluding `tests` isn't sufficient then? I always assumed that this is enough.

setup.cfg
mgorny: You are correct. Updated now.

mgorny:
Apparently the idea is that setuptools does pattern matching of every package against the `exclude` list.

Alternatively, a similar effect can be achieved via:

```
exclude=
    tests*
```

instead of the two entries (because `tests*` matches both `tests` and `tests.myapp`). But it's kinda ugly-ish since it would also match `testsfoo` and stuff like that (not that it's going to exist).

I can adjust it either way, it's a matter of your preference.

matthiask:
Yeah, I somehow assumed that adding `tests` would make `find_packages` not even descend into this package/directory.

I agree with you, `tests` and `tests.*` as separate entries seems safer to me.

PR #797, opened 2021-08-27
Support dark django admin theme

tfeldmann:
Before:

Bildschirmfoto 2021-08-27 um 12 16 48

After:

Bildschirmfoto 2021-08-27 um 12 16 36


OliverSieberling:
Hey there, when using firefox, the dark theme is used even thought firefox isn't in dark mode.SCS
image

tfeldmann:
That's strange! Which OS  / FF version are you on?

matthiask:
Hmm. @OliverSieberling Maybe you're not on Django 3.2? Older Django versions do not support a dark mode yet (I think). This also means that this change introduced a minimum Django version of 3.2 for the draggable MPTT admin for anyone using dark mode.

lukasboehm97:
Hey all, I've got the same issue. using Django 3.2.7 and FF 91.0.2 . It works great with Crome however.

tfeldmann:
I submitted a PR which should fix this problem. Please test 👍 #799

PR #799, opened 2021-09-08
Fix light theme asset path

tfeldmann:
Fixes the problems with the light themes discussed in #797. The file pathes were wrong. Tested in Safari and Firefox.


matthiask:
Oops :) Looks good, thanks!

Issue #779, opened 2021-03-10
ValueError at /admin/{my_app}/{my_node_model}/add/ Cannot call get_descendants on unsaved {my_node_model} instances


cecilia-cobalt:
Hi! I'm new to the mptt package. Ran into the following issue and could you check if my suggested fix make sense or could it be me missing some local configs to make it work as expected?

`Changeform view` is raising exception on an `MPTTModel`. Under my admin view, I used `DraggableMPTTAdmin`. 

```
Traceback (most recent call last):
  File "lib/python3.6/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "lib/python3.6/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/lib/python3.6/contextlib.py", line 52, in inner
    return func(*args, **kwds)
  File "lib/python3.6/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "lib/python3.6/site-packages/django/contrib/admin/options.py", line 614, in wrapper
    return self.admin_site.admin_view(view)(*args, **kwargs)
  File "lib/python3.6/site-packages/django/utils/decorators.py", line 130, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "lib/python3.6/site-packages/django/views/decorators/cache.py", line 44, in _wrapped_view_func
    response = view_func(request, *args, **kwargs)
  File "lib/python3.6/site-packages/django/contrib/admin/sites.py", line 233, in inner
    return view(request, *args, **kwargs)
  File "lib/python3.6/site-packages/django/contrib/admin/options.py", line 1653, in add_view
    return self.changeform_view(request, None, form_url, extra_context)
  File "lib/python3.6/site-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "lib/python3.6/site-packages/django/utils/decorators.py", line 130, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "lib/python3.6/site-packages/django/contrib/admin/options.py", line 1534, in changeform_view
    return self._changeform_view(request, object_id, form_url, extra_context)
  File "lib/python3.6/site-packages/django/contrib/admin/options.py", line 1594, in _changeform_view
    form = ModelForm(initial=initial)
[...]

PR #801, opened 2021-09-23
BUGFIX: Cannot call get_descendants on unsaved models using PostgreSQ…

elarroba:
__BUGFIX__:  Cannot call get_descendants on unsaved models using PostgreSQL databases when adding new models through Django Admin.

A UUID primary key field using a PostgreSQL database will always have a value regardless of the model being saved or not.
Therefore is is necessary to add an additional check "__and not self.instance._state.adding__"  to the init function to make sure there is no ValueError when adding new models using the Django Admin interface.

```python
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        if self.instance and self.instance.pk and not self.instance._state.adding:
            instance = self.instance
            opts = self._meta.model._mptt_meta
            parent_field = self.fields.get(opts.parent_attr)
            if parent_field:
                parent_qs = parent_field.queryset
                parent_qs = parent_qs.exclude(
                    pk__in=instance.get_descendants(include_self=True).values_list(
                        "pk", flat=True
                    )
                )
                parent_field.queryset = parent_qs
```

### DIFF
--- a/mptt/__init__.py
+++ b/mptt/__init__.py
@@ -1,7 +1,7 @@
 import django
 
 
-__version__ = "0.13.3"
+__version__ = "0.13.4"
 VERSION = tuple(__version__.split("."))
 
 if django.VERSION < (3, 2):  # pragma: no cover
--- a/mptt/forms.py
+++ b/mptt/forms.py
@@ -174,7 +174,7 @@ class MPTTAdminForm(forms.ModelForm):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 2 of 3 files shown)


elarroba:
Refer to https://github.com/django-mptt/django-mptt/issues/779#issue-826832449

matthiask:
Thank you!

(Incrementing the version number yourself is a bit unusual.)

elarroba:
@matthiask Thank for approving my pull request and for allowing me to contribute.

matthiask:
@elarroba You're welcome! :)

PR #813, opened 2022-07-29
Raise exception when usage through model_mommy/model_bakery is not explicitly indicated

### LINKED ISSUE #694: Docs: Using MPTT with Model Mommy
When running tests on MPTT models, we were unable to use `model_mommy`, since it randomizes fields on the model including `lft` and `rght`.

As per [this StackOverflow](https://stackoverflow.com/a/32272538) answer, a simple recipe will suffice.

```python
from model_mommy.recipe import Recipe

# Define recipe that avoids setting critical fields
my_model_recipe = Recipe(MyModel, lft=None, rght=None)

# Usage in test:
test_model_instance = my_model_recipe.make()
```
The projects docs should include this fix somewhere, maybe in a Testing section.

PetrDlouhy:
Fix #694 by not allowing usage through `model_mommy`/`model_bakery` unless explicitly indicated by `MPTT_ALLOW_TESTING_GENERATORS` settings.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -25,6 +25,7 @@ Topic docs:
    utilities
    changelog
    technical_details
+   testing
 
 .. toctree::
    :maxdepth: 3
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -1,0 +1,24 @@
+===================
+Testing MPTT models
+===================
+
+
+Using testing generators
+========================
+
+Using testing generators such as ``model_mommy`` or ``model_bakery`` is causing random tree fields values, which can cause unexpected behavior.
+To prevent that the ``django-mptt.MPTTModel`` will throw an Exception if made throught model_mommy/model_bakery in test environment unless
+the ``MPTT_ALLOW_TESTING_GENERATORS`` setting is set to True.
+
+You can set the ``MPTT_ALLOW_TESTING_GENERATORS`` setting to True in your Django testing settings.py file or by the ``@override_settings`` decorator for particular test.
+You would probably also have to use recipe and explicitly set the appropriate fields for the model.
+
+.. code-block:: python
+
+    from django.test import override_settings
+    from baker.recipe import Recipe
+
+    @override_settings(MPTT_ALLOW_TESTING_GENERATORS=True)
+    def test_mptt_allow_testing_generators(self):
+        my_model_recipe = Recipe(MyMPTTModel, lft=None, rght=None)
+        test_model_instance = my_model_recipe.make()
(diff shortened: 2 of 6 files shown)


PetrDlouhy:
This is conceptual proposal how to solve #694. @clarkmoody If you agree with this approach, I will polish this and add also some documentation.

This uses a bit hacky approach to detect how the model was generated through `inspect` module. I don't know if there is any better way to do this.

matthiask:
Thank you. I'm not happy with inspecting frames every time a MPTT model is instantiated. 

I think I'd prefer an addition to the docs, something like "Run Node.objects.rebuild() to fix the MPTT attributes in your testsuite if those attributes may be out of sync.", or maybe people should just use a [different library](https://406.ch/writing/django-tree-queries/) when they don't understand the details of how MPTT works at all. Unfixed (or maybe unfixable) problems with django-mptt exist, but as long as you follow the rules (and avoid concurrent updates) you should be good, generally.

PetrDlouhy:
@matthiask Applications like `django-hordak` are already using MPTT, so people might use it and not be avare of it.

This PR is trying to solve big trap (described in #694) when MPTT model is created by `model_bakery`. I would like to avoid running into this exact issue in the future, because it could cause me several hours once again.

Just writing this in docs is not enough, because people would not remember at the moment they are trying to create some model through `model_bakery`.

I don't like inspecting frames either, but I didn't find any better solution.

PetrDlouhy:
@matthiask I have an idea: If you think that checking frames can be somehow heavy on resources, I can run that condition only if it is called in testing environment (detected by `TESTING = sys.argv[1:2] == ['test']` https://stackoverflow.com/questions/6957016/detect-django-testing-mode).

matthiask:
> @matthiask I have an idea: If you think that checking frames can be somehow heavy on resources, I can run that condition only if it is called in testing environment (detected by `TESTING = sys.argv[1:2] == ['test']` https://stackoverflow.com/questions/6957016/detect-django-testing-mode).

Yes, I think that would be better. Or maybe add a new module-level function `_check_no_testing_generators()` (or something, naming is hard) which is a no-op if `model_mommy` or `model_bakery` isn't importable?

PetrDlouhy:
@matthiask OK, I changed the PR to use both mentioned conditions. I am not sure, if it will work in all testing scenarios, but I am sure, that it will not have negative effect in non-testing environments.

matthiask:
LGTM, thanks!

PetrDlouhy:
Now I have added some docs and I think it is ready for review and pull.

matthiask:
Thanks!

PR #816, opened 2022-09-05
Update mptt_change_list.html to include {% change_list_object_tools %}

paramkpr:
It seems that the admin template admin/mptt-change-list.html currently does not have the {% change_list_object_tools %} tag present, which causes the Add button to not be rendered. This change ensures that the Add button is rendered even when the django admin has some kind of skin added to it (eg. jazzmin). 

https://stackoverflow.com/questions/68722249/django-jazzmin-add-button-disappeared-after-adding-mptt-admin/69230357#69230357

### DIFF
--- a/mptt/templates/admin/mptt_change_list.html
+++ b/mptt/templates/admin/mptt_change_list.html
@@ -2,7 +2,26 @@
 {% load admin_list i18n mptt_admin %}
 
 {% block result_list %}
-    {% if action_form and actions_on_top and cl.show_admin_actions %}{% admin_actions %}{% endif %}
+  <div class="row">
+    <div class="col-12 col-sm-8">
+      {% if action_form and actions_on_top and cl.show_admin_actions %}{% admin_actions %}{% endif %}
+    </div>
+    <div class="col-12 col-sm-4">
+      {% block object-tools %}
+        {% block object-tools-items %}
+          {% change_list_object_tools %}
+        {% endblock %}
+      {% endblock %}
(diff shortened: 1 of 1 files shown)


matthiask (changes requested):
Thank you! 

I think a less intrusive patch may be in order; maybe there's a way to replace the `ChangeList` class instead of having to override the template at all...?

Anyway, the `.col-12`and `col-sm-8` classes do not exist in `django.contrib.admin` so I don't think this is the way to do it?

paramkpr:
Gotcha! Let me look into the possibility of overriding the ChangeList class and get back to you. 
In any case, I'll update the template change so that it doesn't use `.col-12` and `.col-sm-8` or any other non vanilla CSS classes.

matthiask:
Thank you for the report and the PR — the underlying observation (the Add button disappearing in some admin skins) is valid.

The fix here adds Bootstrap grid markup (`col-12 col-sm-8`, etc.) and moves `object-tools` inside the `result_list` block. This would change the page layout in unintended ways for users on the standard Django admin, and ties the template to a specific grid system.

The root cause seems to be that third-party admin themes (like Jazzmin) override `object-tools` in their own `change_list.html`, but since `mptt_change_list.html` extends the stock Django template rather than the theme's template, those overrides don't apply. The most reliable workaround for affected projects is to override `mptt_change_list.html` locally and extend the theme's base change list template instead of Django's.

See also #833 for the project's maintenance status.

PR #817, opened 2022-09-17
Point out unclear maintenance of django-treebeard

timoludwig:
I migrated my project from django-mptt to django-treebeard because of the "unmaintained" warning in the README and the alongside recommendations.

Ironically, treebeard seems kind of dead (the [last release](https://pypi.org/project/django-treebeard/#history) was 1,5y ago, [my PR](https://github.com/django-treebeard/django-treebeard/pull/245) has been waiting for a review for 8 months) and there has been some progress in this repo instead.

So to prevent others from making the same mistake I made, I would point out that migrating the tree library does not necessarily improve the maintenance.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -5,7 +5,7 @@
 Alternatives to django-mptt include:
 
 * `django-treebeard <https://pypi.org/project/django-treebeard/>`_ includes a MPTT
-  implementation (called nested set)
+  implementation (called nested set), but the state of maintenance is unclear.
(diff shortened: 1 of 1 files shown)


matthiask:
Wow that's unfortunate. Thanks for the PR, it's certainly a good idea to point this out. 

I'm a bit surprised though since django-cms uses django-treebeard and I'd think that such a high profile project would keep their dependencies more up to date.

PR #818, opened 2022-09-22
fix memory error with large levels, add tests

### LINKED ISSUE #814: _get_level_indicator() with large level can eat up whole memory
When `TreeNodeChoiceFieldMixin._get_level_indicator()` is called with MPTT `obj` with large `level` (like 963849161 which I got generated by `model_bakery`), the function can get extremely memory hungry and freezes the whole computer until it finally fails with:
```
terminated by signal SIGKILL (Forced quit)
```

I think, that the level needs to be limited to some sane number.
I also think it might be acceptable to raise an exception if the level is larger than limit, that would make debugging the problem easier.

PetrDlouhy:
Fixes #814

### DIFF
--- a/mptt/forms.py
+++ b/mptt/forms.py
@@ -39,7 +39,7 @@ class TreeNodeChoiceFieldMixin:
 
     def _get_level_indicator(self, obj):
         level = self._get_relative_level(obj)
-        return mark_safe(conditional_escape(self.level_indicator) * level)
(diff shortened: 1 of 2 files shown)


PetrDlouhy:
I hope that the failed test with `Network is unreachable` error message would run on the second attempt.

matthiask:
Yeah, the error is certainly unrelated. Thank you!

PR #819, opened 2022-10-06
Update overview.rst

iamjatinyadav:
A small type error remove from overview.rst


matthiask:
Thanks!

The change isn't correct though. The expression is "to whet one's appetite", not "to what one's appetite".

PR #820, opened 2022-11-26
Optimized rebuild method

ipakeev:
Rebuilds trees 5-10 times faster.

### DIFF
--- a/mptt/managers.py
+++ b/mptt/managers.py
@@ -635,7 +635,7 @@ class TreeManager(models.Manager.from_queryset(TreeQuerySet)):
         qs = self._mptt_filter(parent_id__isnull=True, **filters)
         if opts.order_insertion_by:
             qs = qs.order_by(*opts.order_insertion_by)
-        return list(qs.only("id"))
+        return list(qs.only("pk"))
 
     def _get_children(self, **filters):
         opts = self.model._mptt_meta
@@ -644,8 +644,8 @@ class TreeManager(models.Manager.from_queryset(TreeQuerySet)):
             qs = qs.order_by(*opts.order_insertion_by)
 
         children = defaultdict(list)
-        for child in qs.only("id", "parent_id"):
-            children[child.parent_id].append(child)
+        for child in qs.select_related("parent"):
+            children[child.parent.pk].append(child)
         return children
 
     @delegate_manager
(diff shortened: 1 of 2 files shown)


matthiask (commented):
Thanks! I didn't read the code too closely but it does look like it does the right thing.

Is there a reason to add a whole new manager method though? Can't you modify the existing rebuild functionality instead? This would also mean that existing tests (if there are any) can be reused.

ipakeev:
Done 👍. 
I also optimized the partial_rebuild method.

mptt/managers.py
```diff
@@ -619,48 +620,102 @@ def root_nodes(self):
...
             qs = qs.order_by(*opts.order_insertion_by)
-        pks = qs.values_list("pk", flat=True)
 
-        rebuild_helper = self._rebuild_helper
-        idx = 0
-        for pk in pks:
-            idx += 1
-            rebuild_helper(pk, 1, idx)
+        children = defaultdict(list)
+        for child in qs.only("id", "parent_id"):
```
matthiask, line 647: I'm wondering. Is it guaranteed that those two fields are named `id` and `parent_id`? Shouldn't you use `pk` and `_mptt_meta.parent_attr` or something?

matthiask (commented):
Thanks, this is looking good! I have a doubt re. the naming of model fields but apart from that I think it's good to go. Nice to see the improvements in the testsuite.

mptt/managers.py
ipakeev: Yes, that's a good idea.

matthiask:
Thank you!

mptt/managers.py
```diff
@@ -619,48 +620,102 @@ def root_nodes(self):
...
             qs = qs.order_by(*opts.order_insertion_by)
-        pks = qs.values_list("pk", flat=True)
 
-        rebuild_helper = self._rebuild_helper
-        idx = 0
-        for pk in pks:
-            idx += 1
-            rebuild_helper(pk, 1, idx)
+        children = defaultdict(list)
+        for child in qs.select_related("parent"):
```
henrikek, line 647: When I upgrade to version 0.15 i run into problem when the name of related is not "parent". Why not use opts.parent_attr?

PR #824, opened 2023-04-14
Add support for django 4.2

jokiefer:
[index_together](https://docs.djangoproject.com/en/4.2/ref/models/options/#index-together) is deprecated since django 4.2

This PR changes the way how indexes are append to the model. It converts the indexes from the `index_together` attribute to the new [Index](https://docs.djangoproject.com/en/4.2/ref/models/indexes/#index-options) objects to add backward compatibility.

Cause the mptt specific fiels are not available on django checks, i implemented a new MPTTMeta option to provide custom Indexes. This is a bit hacky, cause there is a reason, why django proofs the indexes on that early time.

However, in my pov there is still the need of using this mptt package, cause other alternative packages does not fit all edge cases which are covered by this package. As discussed in https://github.com/matthiask/django-tree-queries/issues/54, it is hard to prefetch all ancestors per node without mptt. You can do it with [django-treebeard](https://pypi.org/project/django-treebeard/). But the nested set implementation of django-treebeard does not adds a parent foreignkey. The foreignkey makes it possible to select the related parent per node.

### DIFF
--- a/mptt/utils.py
+++ b/mptt/utils.py
@@ -5,8 +5,11 @@ trees.
 import copy
 import csv
 import itertools
+import random
+import string
 import sys
 
+from django.db.models.indexes import Index
 from django.utils.translation import gettext as _
 
 
@@ -292,3 +295,38 @@ def get_cached_trees(queryset):
             current_path.append(obj)
 
     return top_nodes
+
+
+def get_random_string(length):
+    # With combination of lower and upper case
+    return "".join(random.choice(string.ascii_letters) for i in range(length))
+
+
+def get_default_index(model_cls):
+    return Index(
+        fields=(model_cls._mptt_meta.tree_id_attr, model_cls._mptt_meta.left_attr),
+        name=f"{get_random_string(12)}_default_mptt_idx",
+    )
+
+
+def convert_index_together(index_together) -> Index:
+    return Index(fields=index_together, name=f"{get_random_string(30)}")
+
+
+def append_indexes(model_cls):
+    default_index = get_default_index(model_cls)
+    indexes = [default_index] + model_cls._mptt_meta.indexes
+    if hasattr(model_cls._meta, "index_together") and model_cls._meta.index_together:
+        for index in model_cls._meta.index_together:
+            indexes.append(convert_index_together(index))
+        model_cls._meta.index_together = []
+
+    for index_to_be_appended in indexes:
+        existing = False
+        for existing_index in model_cls._meta.indexes:
+            if index_to_be_appended.fields == existing_index.fields:
+                existing = True
+                break
+        if not existing:
+            print("appending on", model_cls)
+            model_cls._meta.indexes.append(index_to_be_appended)
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,7 @@
 [tox]
 envlist =
     py{36,37,38,39}-dj{22,30,31}
-    py{38,39,310}-dj{32,40,41,main}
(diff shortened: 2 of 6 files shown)


tox.ini
```diff
@@ -17,6 +17,7 @@ deps =
     dj32: Django>=3.2,<4.0
     dj40: Django>=4.0,<4.1
     dj41: Django>=4.1,<4.2
+    dj42: Django>=4.2,<4.3
```
matthiask, line 20: There won't be a 4.3 according to Django's versioning policy, this should say `>=4.2,<5.0`.

mptt/utils.py
```diff
@@ -292,3 +295,38 @@ def get_cached_trees(queryset):
...
+
+def get_random_string(length):
+    # With combination of lower and upper case
+    return "".join(random.choice(string.ascii_letters) for i in range(length))
+
+
+def get_default_index(model_cls):
+    return Index(
+        fields=(model_cls._mptt_meta.tree_id_attr, model_cls._mptt_meta.left_attr),
+        name=f"{get_random_string(12)}_default_mptt_idx",
```
matthiask, line 308: `get_random_string` already exists in `django.utils.crypto`. But I think it's not a good fit here. Won't this regenerate migrations over and over?

tox.ini
jokiefer: even doesn't know about that policy.

mptt/utils.py
jokiefer: Interesting... i do the migrations for the test app and no `AddIndex` Object is part of the migrations. So currently there are no indexes build from...

But back to the naming problem. The name of an index is limited by 30 characters and needs to be unique. The common way from django is to use [name='%(app_label)s_%(class)s_default_mptt_index')](https://docs.djangoproject.com/en/4.2/ref/models/indexes/#django.db.models.Index.name). But in our test app there will be names which are longer than 30 characters. So i made the workaround with a random string (12 chars) followed by the '_default_mptt_idx'. 

So we got two problems here. 

* No Index objects are part of migrations. For now i don't know why.
* dirty index naming; maybe they will reproduce migrations over and over.

matthiask: Yeah, I think Django has a way of creating reproducible alphanumeric suffixes for some of the database object names where it needs them. I'm not sure how it works (I didn't do any research) but I have seen some of those names in dumps etc.

> Interesting... i do the migrations for the test app and no AddIndex Object is part of the migrations. So currently there are no indexes build from...

Does that mean that the indexes themselves aren't actually created? I'm sorry for the dumb question. 

(I have marked the django-mptt package as unmaintained since maintaining it was honestly thankless, so I'm sorry for not putting more energy into the review. I hope you understand where I'm coming from.)

jokiefer: > Does that mean that the indexes themselves aren't actually created?
Yes; the index operations such as creating altering shall be part of the generated migration files

> (I have marked the django-mptt package as unmaintained since maintaining it was honestly thankless, so I'm sorry for not putting more energy into the review. I hope you understand where I'm coming from.)

I understand that. It is very hard to maintain well. I think there are some overengineered parts such as the renaming mptt attribute stuff. That's also why we can't do this issue easily.

However; I will look again at the django-treebeard package if it is even possible to implement a simple parent foreignkey for me

jokiefer: i started with a new implementation of [django-mptt2](https://github.com/jokiefer/django-mptt2) cause no other project fits my use cases fully.

PR #830, opened 2023-10-05
fix #658 (fix get_cached_trees if there are multiple trees in queryset)

### LINKED ISSUE #658: get_cached_trees - Utility Function appears to not cache parents correctly if passed multiple trees

PetrDlouhy:
I have reworked #731. I have included only the changed `get_cached_trees()` and added some tests to ensure the fixed behavior.


matthiask:
Thanks!

PR #834, opened 2023-12-07
GitHub Actions: Test Python 3.12 and Django 5.0

cclauss:


matthiask:
Thanks! Checks seem to not have run for the last commit, could you try triggering the build again? Maybe by squashing the commits and force-pushing your branch or something.

cclauss:
https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex we needed to drop the `p` at the end as automated in https://docs.astral.sh/ruff/rules/deprecated-unittest-alias

matthiask:
Thanks!

PR #836, opened 2024-01-08
Update manager to support custom parent name

matmair:
Currently a harcoded 'parent' attribute is used for the lookup. If you use the Meta attribute parent_attr (our code base uses this for legacy schema reasons) this breaks.

### DIFF
--- a/mptt/managers.py
+++ b/mptt/managers.py
@@ -646,8 +646,8 @@ class TreeManager(models.Manager.from_queryset(TreeQuerySet)):
             qs = qs.order_by(*opts.order_insertion_by)
 
         children = defaultdict(list)
-        for child in qs.select_related("parent"):
-            children[child.parent.pk].append(child)
+        for child in qs.select_related(opts.parent_attr):
+            children[getattr(child, opts.parent_attr).pk].append(child)
         return children
 
     @delegate_manager


matthiask (commented):
Thanks, looks good. It would be great if you could add or extend a test to avoid regressions in the future?

matmair:
I will look into that in the coming days

harrisonbarlow:
@matmair I was about to submit a PR with your exact changes when I saw this. Happy to lend a hand if you need

matmair:
@harrisonbarlow if you have suggestions how this can be tested in a pythonic way I would be very open to that. All solutions I came up with till now are pretty huge and unwildy

harrisonbarlow:
Hi @matmair

My idea was to create a new model with a custom parent_attr
`class CustomParentAttrModel(MPTTModel):
    name = models.CharField(max_length=50)
    custom_parent = TreeForeignKey(
        "self", null=True, blank=True, related_name="children", on_delete=models.CASCADE
    )

    class MPTTMeta:
        parent_attr = "custom_parent"`

I first noticed this issue when trying to rebuild the tree with a custom parent_attr. So this may provide the coverage we need
`class CustomParentAttrModelTestCase(TreeTestCase):
    def test_rebuild(self):
        root = CustomParentAttrModel.objects.create(name="root")
        child = CustomParentAttrModel.objects.create(name="child", custom_parent=root)
        CustomParentAttrModel.objects.rebuild()

        self.assertTreeEqual(
            CustomParentAttrModel.objects.all(),
            """
            1 - 1 0 1 4
            2 1 1 1 2 3
            """
        )

        child.delete()
        CustomParentAttrModel.objects.rebuild()

        self.assertTreeEqual(
            CustomParentAttrModel.objects.all(),
            """
            1 - 1 0 1 2
            """
        )`

What do you think?

matthiask:
@harrisonbarlow The test looks good to me, something like this is certainly sufficient to avoid regressions.

matmair:
Thank you for the suggestion @harrisonbarlow, I added them

harrisonbarlow:
Thanks @matmair 

@matthiask if you're happy with these changes and the tests are passing, what's the chance of getting a patch release with them?

matthiask:
High, I have just pushed 0.16 to PyPI. Thanks to you all!

PR #841, opened 2025-04-04
Modernize support for Django and Python versions.

felixxm:


matthiask (approved):
Thanks @felixxm 

LGTM. I'm sure you have enough on your plate already but if you would be interested in commit and PyPI access I'm very happy to hand it out.

felixxm:
> Thanks @felixxm
> 
> LGTM. I'm sure you have enough on your plate already but if you would be interested in commit and PyPI access I'm very happy to hand it out.

I will push more commits to this PR with Django 5.1 and 5.2 support in the next few minutes. Unfortunately I cannot promise any time for future maintenance :disappointed: but if you would be so kind and release a new version with Django 5.1 and 5.2 I'd be really grateful :gift_heart:

matthiask:
Thanks, I'm happy to do that! Even though I tried stepping back from this project four years ago. 😅 

The number of sites where I'm still using mptt have dropped below 10, and all of them are in maintenance mode (or worse), so definitely no guarantees from my side either :-) but it's all good.

felixxm:
> Thanks, I'm happy to do that!

Many thanks :green_heart: 

> Thanks, I'm happy to do that! Even though I tried stepping back from this project four years ago. 😅

I know this pain :people_hugging:

felixxm:
Should be ready, we have Django 4.2 -> 5.2 tested with no deprecation warnings or failures.

matthiask:
@felixxm Released as 0.17. Thank you!

PR #844, opened 2025-08-26
Remove duplicate index_together logic in MPTTModelBase

francoisfreitag:
The `index_together` is already added in
https://github.com/django-mptt/django-mptt/blob/755b128657a2ea383282fcb42f0430815a89490d/mptt/models.py#L376-L385, under the guard `is_cls_tree_model`.
No need to repeat it under `if not is_abstract and is_cls_tree_model`.

### DIFF
--- a/mptt/models.py
+++ b/mptt/models.py
@@ -401,15 +401,6 @@ class MPTTModelBase(ModelBase):
                                 tree_manager = cls_manager
                                 break
 
-                if is_cls_tree_model and django.VERSION < (5,):
-                    idx_together = (
-                        cls._mptt_meta.tree_id_attr,
-                        cls._mptt_meta.left_attr,
-                    )
-
-                    if idx_together not in cls._meta.index_together:
-                        cls._meta.index_together += (idx_together,)
-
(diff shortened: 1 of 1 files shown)


matthiask:
I don't really know if we have different code paths which hit one or the other branch. I don't even remember if `mptt.register` still exists? I think that initially django-mptt didn't assume people were deriving from `TreeModel`.

francoisfreitag:
The second branch is more specific (`not is_abstract and is_cls_tree_model`) that the first one (`is_cls_tree_model`), so we can be sure to have entered the first one if we enter the second.

As to whether that is used, a project at work is based on https://github.com/ellmetha/django-machina/, which [builds its `Forum` model as an `MPTTModel`](https://github.com/ellmetha/django-machina/blob/main/machina/apps/forum/abstract_models.py#L31). A colleague noticed the performance issue, because the index on `tree_id, lft` was missing. Analyzing the cause led me here, because that index isn’t created for Django 5+. This will be fixed with #845. In order to simplify #845, this PR removes unnecessary code.

matthiask:
Makes sense, thanks for the explanation!

PR #847, opened 2026-05-26
Add Python 3.14 to the testing

cclauss:
The cron tests are starting to fail on `djmain`: https://github.com/django-mptt/django-mptt/actions

https://docs.djangoproject.com/en/stable/faq/install/#what-python-version-can-i-use-with-django

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,8 +2,8 @@ name: Tests
 
 on:
   push:
-    # branches:
-    #   - main
+    branches:
+      - main
   pull_request:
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -2,8 +2,8 @@ name: Tests
 
 on:
   push:
-    branches:
-      - main
+    # branches:
+    #   - main
```
matthiask:
```suggestion
    branches:
      - main
```

matthiask:
Thanks!

PR #849, opened 2026-06-02
Add 'check' to the 'ruff' command'

cclauss:
Use the linter with `ruff check`
* https://docs.astral.sh/ruff/linter


matthiask:
Thanks!

